PR #5, opened 2012-11-24
cairo backend: MeasureString line wrap fix

eisnerd:
I can produce and attach sample code if needed, but the gist is that CursorX should be updated before WrapX is noted. Otherwise, MaxX may later be updated from WrapX, causing the bounding rectangle given by MeasureString to be too narrow by the width of the last character of the longest wrapped line. If the bounding rectangle is then used in a DrawString, the wrap results can be different between the first MeasureString and the subsequent DrawString, which should not happen if the calculated bounding rectangle is passed on without modification.

There are a number of bugs on the xamarin and novell bugzillas relating to line wrapping with the cairo backend which, though none exactly matching this problem, have been put off with, "this will work with the pango backend." I needed this to work today with the cairo backend on OSX for a text-heavy application, however.


directhex:
build

directhex:
https://jenkins.mono-project.com/job/test-libgdiplus-pull-request/11/

PR #13, opened 2013-11-25
[V2] Fix cross-compilation and drop dependency towards X11

pcercuei:
As required, the three commits with a notice that the code is contributed under the terms of the MIT license.


Arthur2e5:
> @alexrp We should at least print a message letting the user know if we're building with or without X11.

Perhaps some `configure` test for `CAIRO_HAS_XLIB_SURFACE` will be more suitable for this purpose.

akoeplinger:
This was merged as part of https://github.com/mono/libgdiplus/pull/46. Thanks.

PR #34, opened 2015-09-24
correctly resolve the variable of jpeg library prefix

v4hn:
The resulting string is later exported (as part of GDIPLUS_LIBS) to the
generated pkg-config file and this file shouldn't contain
-L${libjpeg_prefix}/lib -ljpeg as libjpeg_prefix is unknown in this context...


directhex:
approve

v4hn:
did that mean "merge"? Because it's still not merged :)

Therzok:
This looks definitely fine to me. Merging.

PR #35, opened 2015-11-02
Feature/conditional x11

haf:
This PR rebases the PR for compiling without x11 support.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -19,8 +19,8 @@ fi
 GLIB_REQUIRED_VERSION="2.2.3"
 PKG_CHECK_MODULES(BASE_DEPENDENCIES, glib-2.0 >= $GLIB_REQUIRED_VERSION)
 
-GDIPLUS_LIBS="`pkg-config --libs glib-2.0 `"
-GDIPLUS_CFLAGS="`pkg-config --cflags glib-2.0 `"
+GDIPLUS_LIBS="`$PKG_CONFIG --libs glib-2.0 `"
+GDIPLUS_CFLAGS="`$PKG_CONFIG --cflags glib-2.0 `"
 
 CAIRO_REQUIRED_VERSION="1.6.4"
 PKG_CHECK_MODULES(CAIRO, cairo >= $CAIRO_REQUIRED_VERSION)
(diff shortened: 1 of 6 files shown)


haf:
approve

akoeplinger:
whitelist

qmfrederik:
@akoeplinger @haf 

I'm working on getting Mono's System.Drawing to work on .NET Core on OS X. This patch would help building libgdiplus on more recent versions of OS X (which don't have X11).

From what I gather, this is a rebased version of #13, which was opened in 2013. Anything I can do to help to have this patch merged?

This would also help Homebrew/legacy-homebrew#45393.

haf:
It works and is ready to merge. It would seem the people behind this repo has taken a hiatus.

qmfrederik:
@haf This fix causes the build to fail on Linux, so I'm guessing it can't go in "as is", I'll comment on the diff itself

src/graphics-private.h
```diff
@@ -105,8 +105,10 @@ typedef struct _Graphics {
 	cairo_t			*ct;
 	GpMatrix		*copy_of_ctm;
 	cairo_matrix_t		previous_matrix;
+#ifdef CAIRO_HAS_XLIB_SURFACE
 	Display			*display;
```
qmfrederik: It looks like Display and Drawable are never used in the codebase (except for setting them to NULL), so I'd guess it'd be better to just remove them alltogether?

src/graphics.c
```diff
@@ -216,7 +218,9 @@ GdipCreateFromHDC (void *hDC, GpGraphics **graphics)
 	cairo_surface_t *surface;
 	int x, y;
 	unsigned int w, h, border_w, depth;
+#ifdef CAIRO_HAS_XLIB_SURFACE
 	Window root;
```
qmfrederik: Same here, looks like this is just an unused variable?

configure.ac
```diff
@@ -363,17 +363,16 @@ else
 fi
 AM_CONDITIONAL(HAVE_LIBEXIF, test x$libexif_pkgconfig = xyes)
 
-dnl
```
qmfrederik: Wouldn't it be better to keep the X11 support by default, and only remove it on Mac?

Mask66:
Conditional X11 would also be a very good thing for headless FreeBSD systems, but I couldn't get the patches  to work , especially  disabling tests won't work for me.

akoeplinger:
This was merged with https://github.com/mono/libgdiplus/pull/46. Thanks.

PR #36, opened 2016-02-17
Remove redundant transformations set on image load which cause failur…

gofman:
Remove redundant transformations set on image load which cause failures with latest libpng

The transformations are set after PNG file is already loaded through png_read_png(). The code below handles actual bits per pixel, and the transformation been set had no effect. But transormations set after image read were causing failures (err msg "invalid after png_start_read_image or png_read_update_info" followed by image load abort) with latest libpng, making PngCodecTest failing in Mono, as well as loading all the images with bpp < 8 and paletted images.

### DIFF
--- a/src/pngcodec.c
+++ b/src/pngcodec.c
@@ -433,25 +433,6 @@ gdip_load_png_image_from_file_or_stream (FILE *fp, GetBytesDelegate getBytesFunc
 
 		interlace = png_get_interlace_type (png_ptr, info_ptr);
 
-		/* NB: none of these png_set_* methods actually do anything after png_read_png() is called! */
-		/* According to the libpng manual, this sequence is equivalent to
-		* using the PNG_TRANSFORM_EXPAND flag in png_read_png. */
-		if (color_type == PNG_COLOR_TYPE_PALETTE) {
-			png_set_palette_to_rgb (png_ptr);
-		}
-
-		if ((color_type == PNG_COLOR_TYPE_GRAY) && (bit_depth < 8)) {
-#if PNG_LIBPNG_VER > 10399
-			png_set_expand_gray_1_2_4_to_8 (png_ptr);
-#else
-			png_set_gray_1_2_4_to_8(png_ptr);
-#endif
-		}
-
-		if (png_get_valid(png_ptr, info_ptr, PNG_INFO_tRNS)) {
-			png_set_tRNS_to_alpha(png_ptr);
-		}
-
 		stride = (width * 4);
 		gdip_align_stride (stride);
 


gofman:
.. If there was a need to perform these transformations on PNG load it could be done through the flags to png_read_png, but this would break functionality (that's what I've tried initially but got PngCodecTest test failing with wrong bitmap format resulting ifrom GdiPlus).

qmfrederik:
@gofman if you're still willing to move this forward, do you have a simple test case that illustrates the issue you're fixing with this PR?

gofman:
I just reproduced the issue on B&W png image, I am attaching it. If to replace test.png image with this one in tests and run testgdi, without my patch it shows an error on loading png file: "libpng error: invalid after png_start_read_image or png_read_update_info", image is not displayed. It is fixed by the patch. I just tested that with libpng 1.6.27 (the error shows up starting from some version of libpng, quite an old one though).
It's been a while since I investigated this issue, but as I recall now this code I am removing was redundant even in older versions of libpng (as this attributes cannot be set at this stage of reading png file), but older versions simply ignored that while newer give an error. The bitmaps are actually loaded (converted) by existing code without the requested expansion to 8 bit.


qq

gofman:
I rebased my fork to the current upstream. There is a commit by Miguel de Icaza merged after my original pull request (4ae938972 'Fix libgdiplus to handle 2bpp color PNG files') which, in particular, adds a comment above the lines I am removing: "/* NB: none of these png_set_* methods actually do anything after png_read_png() is called! */". The problem is that with the (relatively) new libpng versions these png_set... actually break the thing causing a fatal error.

qmfrederik:
@gofman Thanks for the prompt reply! I had another look into this. Yes, it seems the code never worked; if it breaks (because the call to `png_set_*` comes after `png_read_png`, which is invalid, and newer versions of libpng are more strict about this), removing that block already helps.

The original intent was to expand PNG images with a pixel depth of 2 bit to 8 bit, because that's what GDI+ does as well. I don't know why GDI+ does that, and I think deviating from the GDI+ behavior is better than crashing.

If anyone ever wants to match the GDI+ behavior, the easiest way to do that seems to be to pass `PNG_TRANSFORM_EXPAND` to `png_read_png`, but only when the pixel depth is exactly 2.

To know the pixel depth, you'd need to call `png_read_info`; but I don't think you can call `png_read_png` after that, so you'd need to close and reopen the file. That can be tricky when you're reading from a stream (because you would need to seek to the beginning of the file, and the stream may be forward-only).

@akoeplinger LGTM, would look even better with a unit test. Let me see if I can do something about that.

gofman:
@qmfrederik Regarding the unit test, there was already a unit test in main Mono tree (mcs/class/System.Drawing/Test/System.Drawing) which was already failing for me with Mono/gdiplus built with current libpng, while this change fixed it. I could possibly put an extra test to gdiplus test here in libgdiplus if it makes sense, but the problem this patch is fixing still will be reproducable with the new libpng only.

akoeplinger:
Makes sense, thanks!

PR #37, opened 2016-02-19
patch from Romain Tartiere

9999-ebuilds:
see https://bugs.funtoo.org/browse/FL-816
and https://github.com/gentoo/dotnet/issues/173

../src/.libs/libgdiplus.so: undefined reference to `QuantizeBuffer'
QuantizeBuffer has been removed since giflib 4.2.0
FreeBSD developer Romain Tartiere wrote a patch, which adds QuantizeBuffer to libgdiplus code.

<a href="https://www.freshports.org/commit.php?category=x11-toolkits&port=libgdiplus&files=yes&message_id=201206032007.q53K7Kdp081339@repoman.freebsd.org">2012-06-03, Romain Tartiere</a>:

"Fix build after giflib update.

QuantizeBuffer() is <b>no more public</b> in the giflib but <b>still present</b> in the code.
There is no update (yet) code in the libgdiplus code so copy the relevant lines
from giflib for now."

### DIFF
--- a/src/gifcodec.c
+++ b/src/gifcodec.c
@@ -39,6 +39,292 @@ GUID gdip_gif_image_format_guid = {0xb96b3cb0U, 0x0728U, 0x11d3U, {0x9d, 0x7b, 0
 
 #include "gifcodec.h"
 
+#define COLOR_ARRAY_SIZE 32768
+#define BITS_PER_PRIM_COLOR 5
+#define MAX_PRIM_COLOR      0x1f
+
+static int SortRGBAxis;
+
+typedef struct QuantizedColorType {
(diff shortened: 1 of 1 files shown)


akoeplinger:
We can't take the code from giflib directly due to licensing reasons. Given that this API made a comeback in giflib 5 and most distros/OSes either stuck with 4.1 or moved to 5.0 it shouldn't be too common in the future.

For the remaining cases with giflib 4.2 I've added a stub instead which simply does nothing and prints a warning: https://github.com/mono/libgdiplus/commit/aeccb519b6bd5b42377021cc665cce03281c21ff

Thanks!

PR #41, opened 2016-09-08
Preserve bitmap resolution values for clones

StephenMcConnel:
This fixes https://bugzilla.xamarin.com/show_bug.cgi?id=44127.


src/bitmap.c
```diff
@@ -1065,8 +1065,6 @@ GdipCloneBitmapAreaI (int x, int y, int width, int height, PixelFormat format,
 					  GpBitmap *original, GpBitmap **bitmap)
 {
 	GpBitmap	*result;
-	FrameData	*frame;
-	BitmapData	*bitmap_data;
```
StephenMcConnel: These variables aren't used in the method, so I deleted them while here.

PR #43, opened 2016-10-26
Do not zero RGB pixels if A == 0

SirCmpwn:
This fixes a problem introduced by 5ed200d5d73edc54969b34c708d18728d38c070b.

### DIFF
--- a/src/pngcodec.c
+++ b/src/pngcodec.c
@@ -465,16 +465,12 @@ gdip_load_png_image_from_file_or_stream (FILE *fp, GetBytesDelegate getBytesFunc
 				for (i = 0; i < height; i++) {
 					png_bytep rowp = row_pointers[i];
 					for (j = 0; j < width; j++) {
+						BYTE b = rowp[2];
+						BYTE g = rowp[1];
+						BYTE r = rowp[0];
 						BYTE a = rowp[3];
-						if (a == 0) {
-							set_pixel_bgra (rawptr, 0, 0, 0, 0, 0);
-						} else {
-							BYTE b = rowp[2];
-							BYTE g = rowp[1];
-							BYTE r = rowp[0];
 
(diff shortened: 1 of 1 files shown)


SirCmpwn:
Reached out to Sebastien via email for comment on the change this reverts.

peppy:
Just a few words to add here as to what we've noticed when using `System.Drawing` and the motivation for this change:

When uploading textures to GL and then either mipmapping or scaling below native resolution, GL will average neighbouring pixels. For this reason, for cases like text rendering we regularly use textures which have RGB set to 255 (or a similar non-zero value) and only alpha channel adjusted.

The current PNG load method discards this colour data when the alpha value is zero, which mean that when adjacent pixels (or A==0 and A>0) are averaged, rather than the RGB values being identical and thus unaffected, we are suddenly averaging with RGB (0,0,0).

Microsoft's `System.Drawing` implementation produces behaviour identical (incorrect by PNG spec) to what libgdiplus does in the case of

`new Bitmap(Image.FromFile(@"image.png"))`

where it seems to composite the png into the bitmap as an extra step.

The following two methods provide differing (correct by PNG spec) behaviour from libgdiplus. Microsoft's `System.Drawing` loads pixels as they are represented by the PNG - including RGB data when A==0.

`new Bitmap(Stream stream)`
`new Bitmap(string filename)`

which is what we are hoping to see fixed.

Not sure if it's worth keeping the "broken" behaviour for the first case or not; that depends on what the goals of libgdiplus are and if you want to be correct to the MS interpretation or the PNG specification.

SirCmpwn:
Note: email to sebastian bounced.

akoeplinger:
build

akoeplinger:
I think this is a reasonable change, thanks!

SirCmpwn:
Thanks!

PR #44, opened 2017-01-18
Include pkg-config check for libpng16

hasufell:
Otherwise, weird side-effects can occur, such as libgdiplus
linking to multiple png versions, when e.g. png-1.6 and png-1.2 are
both available.

Ultimately this should be fixed more generic and use something like `PKG_CHECK_MODULES([PNG], [libpng >= 1.6])`, picking the _latest_ libpng always. Also see the [pkg-config manpage](https://linux.die.net/man/1/pkg-config) explaining those proper macros. For now, I just kept the logic.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -313,41 +313,49 @@ else
 fi
 
 dnl Test for libpng
-    AC_MSG_CHECKING(for libpng14)
-    if $PKG_CONFIG --exists libpng14 ; then
+    AC_MSG_CHECKING(for libpng16)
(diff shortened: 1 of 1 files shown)


akoeplinger:
Looks good to me, though I'll need to add a small tweak to make sure we're still using libpng14 on OSX until we can upgrade the version of libpng we ship with the bockbuild packaging tool (as this change would cause libpng16 to be picked up e.g. from brew).

Thank you!

PR #47, opened 2017-03-19
Emit tRNS PNG chunk for images with alpha channel in palette.

filipnavara:


filipnavara:
@akoeplinger Any chance you could look into it since you were the last one to commit to this repository and you have some work pending on ResGen in mono?

Our PNG images referenced by .resx files get corrupted by the System.Drawing code when serialising them into binary form. Microsoft implementation preserves the exact byte-to-byte representation of the images, while mono/System.Drawing+libgdiplus doesn't. We need to ensure that the generated images are at least visually same as the original, which is what this patch is trying to fix.

qmfrederik:
@filipnavara This looks good to me; would you be able to write a simple unit test for this? It could be e.g. a C# unit test which takes one of your images, serializes it to its binary form using System.Drawing, and then makes sure the alpha channel is serialized correctly, perhaps by opening the serialized image again?

The test should fail on the current implementation of libgdiplus and pass on .NET on Windows.

filipnavara:
Sure, I can provide a test case. I originally had a test for the behaviour directly in libgdiplus, but I didn't submit it because I was afraid it would break compilation on some of the supported platforms. If a System.Drawing test case is the preferred option I will do that.

akoeplinger:
@filipnavara tests shouldn't break compilation as running tests is a separate step :)

If you have a unit test that works directly in libgdiplus that'd be preferable as it's closer to the source.

filipnavara:
Here's the test case. Please let me know if it is satisfactory.

akoeplinger:
Looks great, thank you!

PR #48, opened 2017-04-26
Visual C++ support: Define API visibility

qmfrederik:
When compiling libgdiplus with Visual C++, all functions would be marked as internal because they lack the `__declspec( dllexport )` attribute.

To work around this, the `WINGDIPAPI` is defined symbol (the same symbol which is used in the public GDI+ API, see [MSDN](https://msdn.microsoft.com/en-us/library/windows/desktop/ms533977(v=vs.85).aspx)), which is left empty when consuming the public API and when on non-msvc compilers, and set to `__declspec( dllexport )` when compiling with msvc.


qmfrederik:
@akoeplinger This is part of a larger effort on my side to get Mono's libgdiplus up and running on Windows. It's main purpose is to make debugging the native code bit easier (using Visual Studio on Windows), but as a side effect, it also brings System.Drawing on Windows platforms which ship without GDI+ (Nano Server).

The wider work is here: https://github.com/CoreCompat/libgdiplus/tree/features/msvc. This PR Is the largest individual change needed to make it work; if this is something you want to ingest I'll also create PRs for the other, smaller changes required.

akoeplinger:
LGTM. The commit message mentions GDIP_EXTERNAL instead of WINGDIPAPI, please fix that.

I don't have anything against getting the msvc fixes into the repo :)

akoeplinger:
In general please make sure the commit message mentions the PR description, it's easier to do code archaeology in the future if it's part of the git history :)

qmfrederik:
Sure, makes sense, fixed it.

PR #49, opened 2017-04-26
Support 48/64bpp PNG images

qmfrederik:
Mono's System.Drawing currently does not support PNG images with 16 bits per pixel - in .NET, RGB colors are usually expressed as an 8-bit value.

Changes the call to png_read_png and passes PNG_TRANSFORM_STRIP_16, which reduces the color depth from 16 bit to 8 bit by dropping the last 8 bits - so you end up with a 24bpp instead of a 48bpp PNG image for RGB images and 32bpp instead of 64bpp for RGBA images.
Net impact is that the quality of the image is slightly reduced instead of refusing to process the image (and potentially crashing the application) altogether; proper support would mean supporting 16-bit color channels.

Partially fixes http://bugzilla.ximian.com/show_bug.cgi?id=80693

### DIFF
--- a/src/pngcodec.c
+++ b/src/pngcodec.c
@@ -279,7 +279,13 @@ gdip_load_png_image_from_file_or_stream (FILE *fp, GetBytesDelegate getBytesFunc
 		png_set_read_fn (png_ptr, getBytesFunc, _gdip_png_stream_read_data);
 	}
 
-	png_read_png (png_ptr, info_ptr, 0, NULL);
+	/* Pass PNG_TRANSFORM_STRIP_16, which basically reduces the color palette from 16-bits to 8-bits
+	 * for 16-bit color depths. The current implementation of libgdiplus doesn't handle bit depths > 8,
+	 * so this acts as a workaround. Net impact is that the quality of the image is slightly reduced instead
+	 * of refusing to process the image (and potentially crashing the application) altogether;
+	 * proper support would mean supporting 16-bit color channels.
+	 * Partially fixes http://bugzilla.ximian.com/show_bug.cgi?id=80693 */
+	png_read_png (png_ptr, info_ptr, PNG_TRANSFORM_STRIP_16, NULL);
 
 	bit_depth = png_get_bit_depth (png_ptr, info_ptr);
 	channels = png_get_channels (png_ptr, info_ptr);


src/pngcodec.c
```diff
@@ -279,7 +279,7 @@ gdip_load_png_image_from_file_or_stream (FILE *fp, GetBytesDelegate getBytesFunc
 		png_set_read_fn (png_ptr, getBytesFunc, _gdip_png_stream_read_data);
 	}
 
-	png_read_png (png_ptr, info_ptr, 0, NULL);
+	png_read_png (png_ptr, info_ptr, PNG_TRANSFORM_STRIP_16, NULL);
```
akoeplinger: please add a short comment above with a short description of why we're doing this and e.g. mention it's a partial fix for bug XXX.

And also add the full description from the PR to the commit message.

qmfrederik:
@akoepling addressed the feedback, let me know if you need anything else. When this and #48 get merged, I'll rebase my branch and open PR(s) for the rest of the changes.

akoeplinger:
Did you run the System.Drawing/WinForms tests against this?

akoeplinger:
Also, while you're in PNG-land, what do you think about #47 and #36 ?

qmfrederik:
I've created a unit test for this fix (will create a separate PR for that in mono), but didn't run the full regression tests yet. I'll do that and keep you posted on the outcome.

#47 LGTM. Would look even better with a unit test.
#36 need to dive deeper into that. The comment in the code, 
> According to the libpng manual, this sequence is equivalent to using the PNG_TRANSFORM_EXPAND flag in png_read_png.

is correct, but the libpng manual also mentions that this code should not work (as mentioned in the PR, you cannot apply transformations after having read the data). So I'm thinking there perhaps was a very specific reason to do this after all. I'll leave a comment there.

qmfrederik:
@akoeplinger I ran the tests on Travis and they all came back green, so we should be good here:
https://travis-ci.org/CoreCompat/libgdiplus/jobs/226503186

PS: I've generated a [`.travis.yml`](https://github.com/CoreCompat/libgdiplus/blob/features/png48bpp-travis/.travis.yml) file which can run System.Drawing and System.Windows.Forms tests on an updated version of libgdiplus in Travis. Let me know if that's something you could host in this repo.

Thoughts on how to just compile the System.Drawing and System.Windows.Forms tests without having to re-compile all of Mono are also welcome :)

akoeplinger:
@qmfrederik 

> PS: I've generated a .travis.yml file which can run System.Drawing and System.Windows.Forms tests on an updated version of libgdiplus in Travis. Let me know if that's something you could host in this repo.

Yes please, that'd be awesome. I thought about this for a while but didn't get around to it.

> Thoughts on how to just compile the System.Drawing and System.Windows.Forms tests without having to re-compile all of Mono are also welcome :)

Hmm, it's tricky since you need to run it against some version of Mono/System.Drawing anyway and I'm not sure doing that with the stable vs. latest master is necessarily better. It seems to compile reasonably fast on Travis so I don't really care :)

PR #50, opened 2017-04-27
Visual C++ compatibility fixes

qmfrederik:
Enables building libgdiplus using Visual C++:

A couple of minor fixes which allow building libgdiplus natively on Visual C++ (Windows)]
- use `tmpfile` instead of `mkstemp` to get generate a temporary file
- Define `_USE_MATH_DEFINES` so `M_PI` is defined
- Don't `#include` Unix-only headers
- Don't redefine `max` and `min`
- Fix macro expansion

### DIFF
--- a/src/font.c
+++ b/src/font.c
@@ -24,6 +24,10 @@
  *	Jeffrey Stedfast <fejj@novell.com>
  */
 
+#ifdef WIN32
+#include "win32_io.h"
+#endif
+
 #include <cairo-features.h>
 #include "gdiplus-private.h"
 #include "font-private.h"
@@ -1074,9 +1078,13 @@ GdipPrivateAddMemoryFont(GpFontCollection *fontCollection, GDIPCONST void *memor
 	if (!memory)
 		return InvalidParameter;
 
+#ifdef WIN32
+	f = CreateTempFile (fontfile);
+#else
 	strcpy((char *) fontfile, "/tmp/ffXXXXXX");
-
 	f = mkstemp((char*)fontfile);
+#endif
+
 	if (f == -1)
 		return FileNotFound;
 
--- a/src/pngcodec.c
+++ b/src/pngcodec.c
@@ -364,15 +364,18 @@ gdip_load_png_image_from_file_or_stream (FILE *fp, GetBytesDelegate getBytesFunc
 				}
 
 				for (i=0; i < num_trans; i++) {
+					png_bytep alpha =
+#if (PNG_LIBPNG_VER > 10399)
+						trans_alpha[i];
+#else
+						info_ptr->trans[i];
+#endif
+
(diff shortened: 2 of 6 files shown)


qmfrederik:
Running the System.Drawing and System.Windows.Forms unit tests on Linux as part of https://travis-ci.org/CoreCompat/libgdiplus/jobs/226520941

qmfrederik:
@akoeplinger System.Drawing + System.Windows.Forms tests are green for this change as well. The pointer-related changes caused a regression, so I'll handle that in a separate PR.

src/font.c
```diff
@@ -1074,9 +1074,14 @@ GdipPrivateAddMemoryFont(GpFontCollection *fontCollection, GDIPCONST void *memor
 	if (!memory)
 		return InvalidParameter;
 
+#ifdef WIN32
+	f = tmpfile();
+#else
 	strcpy((char *) fontfile, "/tmp/ffXXXXXX");
 
 	f = mkstemp((char*)fontfile);
+#endif
```
akoeplinger: `tmpfile()` is not the correct replacement since it'll try to create the file at the system root which won't work if the user doesn't have permissions. Additionally, `fontfile` is not set but we're using it in FcConfigAppFontAddFile below.

It's better to use `GetTempFileName()` + `GetTempPath()`.

src/pngcodec.c
```diff
@@ -364,15 +364,18 @@ gdip_load_png_image_from_file_or_stream (FILE *fp, GetBytesDelegate getBytesFunc
 				}
 
 				for (i=0; i < num_trans; i++) {
+					png_bytep trans =
+#if (PNG_LIBPNG_VER > 10399)
+						trans_alpha[i]; /* alpha */
+#else
+						info_ptr->trans[i]; /* alpha */
+#endif
```
akoeplinger: Let's name the variable `alpha` instead of `trans` and get rid of the comment.

qmfrederik:
@akoeplinger This should be better :). I can't include `windows.h` in `font.c` because libgdiplus redefines a lot of Windows-related functions, so the best option appeared to be to define a mkstemp compat shim in a separate file.

src/font.c
```diff
@@ -1074,9 +1078,12 @@ GdipPrivateAddMemoryFont(GpFontCollection *fontCollection, GDIPCONST void *memor
 	if (!memory)
 		return InvalidParameter;
 
+#ifndef WIN32
 	strcpy((char *) fontfile, "/tmp/ffXXXXXX");
+#endif
 
 	f = mkstemp((char*)fontfile);
```
akoeplinger: I'd like to avoid calling the Win32 reimplementation "mkstemp" when it doesn't follow the semantics of mkstemp. 

Let's call it e.g. `CreateTempFile()` or something like that.

PR #51, opened 2017-04-28
Add Travis file

qmfrederik:
Build libgdiplus on Linux and macOS, and run the System.Drawing and System.Windows.Forms unit tests using the updated libgdiplus.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,24 +11,26 @@ install:
 - if [ "$TRAVIS_OS_NAME" = "linux" ]; then sudo apt-get install -y git autoconf libtool automake build-essential gettext cmake; fi
 
 script:
-# Build libgdiplus
+- echo "travis_fold:start:build_libgdiplus"
 - if [ "$TRAVIS_OS_NAME" = "linux" ]; then ./autogen.sh --prefix=/usr/local/; fi
 - if [ "$TRAVIS_OS_NAME" = "osx" ]; then ./autogen.sh --without-x11 --prefix=/usr/local/; fi
 - make -j4
+- echo "travis_fold:end:build_libgdiplus"
 - make check
 - sudo make install
 
-# Build Mono
+- echo "travis_fold:start:build_mono"
 - cd ..
 - git clone --depth 1 https://github.com/mono/mono
 - cd mono
 - ./autogen.sh --prefix=/usr/local/
-- make -j4 > /dev/null
+- make -j4
+- make -C mcs/class/System.Drawing test
+- make -C mcs/class/System.Windows.Forms test
+- echo "travis_fold:end:build_mono"
 
 # Run the System.Drawing and System.Windows.Forms tests.
 # System.Windows.Forms test don't currently pass on macOS.
 - export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/lib/
-- cd mcs/class/System.Drawing
-- make check
-- cd ../System.Windows.Forms
-- if [ "$TRAVIS_OS_NAME" = "linux" ]; then xvfb-run make check; fi
+- make -C mcs/class/System.Drawing run-test
+- if [ "$TRAVIS_OS_NAME" = "linux" ]; then xvfb-run make -C mcs/class/System.Windows.Forms run-test; fi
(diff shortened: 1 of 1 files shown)


qmfrederik:
@akoeplinger here you go :)

.travis.yml
```diff
@@ -0,0 +1,37 @@
+dist: trusty
+sudo: true
```
akoeplinger: should be `sudo: required` according to Travis docs (yeah it may still work with true but better stick to the docs).

In fact, I think we don't even need `sudo`, we can install mono/libgdiplus into a writable prefix like /tmp and I the other packages should be available via https://docs.travis-ci.com/user/installing-dependencies/#Installing-Packages-with-the-APT-Addon

.travis.yml
```diff
@@ -0,0 +1,37 @@
...
+
+matrix:
+  include:
+    - os: osx
+      env: TARGET=x86_64-apple-darwin RID=osx-x64
```
akoeplinger: we don't need this

.travis.yml
```diff
@@ -0,0 +1,37 @@
...
+    - os: linux
+      env: TARGET=x86_64-unknown-linux-gnu RID=linux-x64
+
+install:
+- if [ "$TRAVIS_OS_NAME" = "osx" ]; then brew install glib cairo libexif; fi
+- if [ "$TRAVIS_OS_NAME" = "linux" ]; then sudo apt-get install -y libgif-dev xvfb; fi
+- if [ "$TRAVIS_OS_NAME" = "linux" ]; then sudo apt-get install -y git autoconf libtool automake build-essential mono-devel gettext cmake; fi
```
akoeplinger: you can remove `mono-devel`. Mono will download a required minimal bootstrapper during build.

This also allows us to make sure we don't have any distro-provided libgdiplus on the system

.travis.yml
```diff
@@ -0,0 +1,37 @@
...
+
+script:
+# Build libgdiplus
+- cd $HOME/build/CoreCompat/libgdiplus
+- if [ "$TRAVIS_OS_NAME" = "linux" ]; then ./autogen.sh --prefix=/usr/; fi
+- if [ "$TRAVIS_OS_NAME" = "osx" ]; then ./autogen.sh --without-x11 --prefix=/usr/; fi
+- make
```
akoeplinger: use `make -j4` to speed up the build

.travis.yml
```diff
@@ -0,0 +1,37 @@
...
+- make check
+- sudo make install
+
+# Build Mono
+- cd $HOME/build/
+- git clone --depth 1 https://github.com/mono/mono
+- cd $HOME/build/mono
+- ./autogen.sh
+- make > /dev/null
```
akoeplinger: use `make -j4` to speed up the build

.travis.yml
```diff
@@ -0,0 +1,37 @@
...
+- if [ "$TRAVIS_OS_NAME" = "osx" ]; then ./autogen.sh --without-x11 --prefix=/usr/; fi
+- make
+- make check
+- sudo make install
+
+# Build Mono
+- cd $HOME/build/
+- git clone --depth 1 https://github.com/mono/mono
+- cd $HOME/build/mono
+- ./autogen.sh
```
akoeplinger: specify `--prefix=/usr/` to be consistent with libgdiplus

qmfrederik:
@akoeplinger Looks like the OS X build is too slow for the tests to run. Unless there's something we can do to reduce the build time, I'd say we just run make and make check for libgdiplus on macOS and skip the Mono tests? (It's already an improvement over what we have now)

akoeplinger:
that's weird, I remember building Mono on Travis on OSX about three years ago and that worked. I pushed a commit that removes the logging to /dev/null so we see more of what's going on. If that doesn't work then I agree we should remove OSX for now :)

qmfrederik:
@akoeplinger Very weird, the build now times out on OS X. I restarted the build today (in CoreCompat) and get the same timeout behavior.

I don't think it's related, but I removed the folds just to be sure.

akoeplinger:
Thanks, this is awesome!

akoeplinger:
I just found out why the OSX build hung: https://github.com/mono/mono/pull/4792 😄

qmfrederik:
@akoeplinger Cool, the OS X build no longer hangs but still times out, unfortunately:

https://travis-ci.org/CoreCompat/libgdiplus/jobs/229998135

PR #52, opened 2017-04-29
Visual C++ Compatibility: Use BYTE* instead of void*

qmfrederik:
`sizeof(void*)` is defined by gcc only, and Visual C++ will not let you increment the value of `void*`-typed pointers. To enable compilation with Visual C++, use the underlying data type - which is `BYTE *`.


qmfrederik:
@akoeplinger This is the last code-level change required to support msvc. I've submitted the vcxproj on a separate PR.

PR #53, opened 2017-04-29
Add Visual C++ project, AppVeyor script

qmfrederik:
This completes the initial work to get libgdiplus compiling with Visual C++ and running on Windows.

Adds a Visual C++ project (and VS solution) which compiles libgdiplus. Most of the dependencies are sourced from HexChat's gtk-win32; others come from CoApp packages on NuGet.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,25 @@
+platform:
+  - Win32
+  - x64
+
+configuration:
+  - Debug
+  - Release
+
+environment:
+  LIBGDIPLUS_VERSION_PREFIX: 4.2
+
+before_build:
+  - cmd: choco install -y wget
+  - cmd: wget "https://dl.hexchat.net/gtk-win32/vc14/x86/gtk-Win32.7z" -O gtk-Win32.7z
+  - cmd: 7z x gtk-Win32.7z -ogtk
+  - cmd: wget "https://dl.hexchat.net/gtk-win32/vc14/x64/gtk-x64.7z" -O gtk-x64.7z
+  - cmd: 7z x gtk-x64.7z -ogtk
+  - cmd: nuget restore
+
+build:
+  project: libgdiplus.sln
+
+on_success:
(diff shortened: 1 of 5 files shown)


qmfrederik:
@akoeplinger This gets the msvc build up and running and adds an AppVeyor script for making sure there are no compile-time regressions.

This was "good enough" for me to develop and test #49 entirely on Windows, so I'm thinking it's worth submitting this as a PR already.

There are a couple of gotchas, though:

- On x86, I the calling convention is off - msvc libgdiplus uses cdecl instead of stdcall; but compiling with sdtcall is not straightforward because the dependencies use cdecl
- On x64, initialization fails because of a bug in the fontconfig code in gtk-win32; I worked around this on my build by just commenting out fontconfig initialization.

If you know of other projects from where I could source those dependencies, let me know.

appveyor.yml
```diff
@@ -0,0 +1,25 @@
...
+
+environment:
+  LIBGDIPLUS_VERSION_PREFIX: 4.2
+
+before_build:
+  - cmd: choco install -y wget
+  - cmd: wget "https://dl.hexchat.net/gtk-win32/vc14/x86/gtk-Win32.7z" -O gtk-Win32.7z
+  - cmd: 7z x gtk-Win32.7z -ogtk
+  - cmd: wget "https://dl.hexchat.net/gtk-win32/vc14/x64/gtk-x64.7z" -O gtk-x64.7z
+  - cmd: 7z x gtk-x64.7z -ogtk
```
akoeplinger: why do we need gtk?

qmfrederik: We don't need GTK. GTK shares dependencies with libgdiplus (cairo, fontconfig, libpng,...), and gtk-win32 provides read-to-use win32 distribution of those dependencies.

If there's another source we can pull those dependencies from, happy to hear. Normally I try to use CoApp packages on NuGet but they are either really old or absent.

migueldeicaza:
Thanks for the responses.   Merging

PR #54, opened 2017-06-26
Fix null pointer dereference in GdipPrivateAddMemoryFont

hughbe:
This fixes the following C# tests (see https://github.com/dotnet/corefx/pull/21552)

```cs
[Fact]
public void AddMemoryFont_ZeroMemory_ThrowsArgumentException()
{
    using (var fontCollection = new PrivateFontCollection())
    {
        AssertExtensions.Throws<ArgumentException>(null, () => fontCollection.AddMemoryFont(IntPtr.Zero, 100));
    }
}

[Theory]
[InlineData(0)]
[InlineData(-1)]
public void AddMemoryFont_InvalidLength_ThrowsArgumentException(int length)
{
    using (var fontCollection = new PrivateFontCollection())
    {
        byte[] data = File.ReadAllBytes(Helpers.GetTestBitmapPath("CodeNewRoman.otf"));

        IntPtr fontBuffer = Marshal.AllocCoTaskMem(data.Length);
        try
        {
            Marshal.Copy(data, 0, fontBuffer, data.Length);
            AssertExtensions.Throws<ArgumentException>(null, () => fontCollection.AddMemoryFont(fontBuffer, length));
        }
        finally
        {
            Marshal.FreeCoTaskMem(fontBuffer);
        }
    }
}

[Fact]
public void AddMemoryFont_Disposed_ThrowsArgumentException()
{
    var fontCollection = new PrivateFontCollection();
    fontCollection.Dispose();

    AssertExtensions.Throws<ArgumentException>(null, () => fontCollection.AddMemoryFont((IntPtr)10, 100));
}
```

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -17,7 +17,7 @@ LDADDS =					\
 	-lm
 
 noinst_PROGRAMS =			\
-	testbits testclip testreversepath testpng
+	testbits testclip testfont testreversepath testpng
 
 if HAS_X11
 noinst_PROGRAMS =
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -42,6 +42,12 @@ testclip_SOURCES =	\
 testclip_DEPENDENCIES = $(TEST_DEPS)
 testclip_LDADD = $(LDADDS)
 
+testfonts_DEPENDENCIES = $(TEST_DEPS)
+testfonts_LDADD = $(LDADDS)
+
+testfonts_SOURCES =		\
+	testfonts.c
+
 testreversepath_SOURCES =	\
 	testreversepath.c
 
@@ -58,12 +64,14 @@ EXTRA_DIST =			\
 	$(testgdi_SOURCES)	\
 	$(testbits_SOURCES)	\
 	$(testclip_SOURCES)	\
+	$(testfonts_SOURCES)	\
 	$(testpng_SOURCES)	\
 	$(testreversepath_SOURCES)
 
 TESTS = \
 	testbits \
 	testclip \
+	testfont \
 	testreversepath \
 	testpng \
 	$(NULL)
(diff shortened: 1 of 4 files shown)


hughbe:
Confirmed, the tests pass on Windows

tests/Makefile.am
```diff
@@ -58,12 +64,14 @@ EXTRA_DIST =			\
...
 	$(testbits_SOURCES)	\
 	$(testclip_SOURCES)	\
+	$(testfonts_SOURCES)	\
 	$(testpng_SOURCES)	\
 	$(testreversepath_SOURCES)
 
 TESTS = \
 	testbits \
 	testclip \
+	testfont \
```
akoeplinger: should be `testfonts`

tests/testfont.c
```diff
@@ -0,0 +1,55 @@
+#ifdef WIN32
```
akoeplinger: file should be called `testfonts.c` to be consistent with what you added in Makefile.am

akoeplinger:
whitelist

tests/testfont.c
hughbe: I think it makes more sense to keep the file singular and just update the makefile to be singular too

akoeplinger:
Seems the Travis build fails for some reason on Linux:

```
testfont.c:8:25: fatal error: GdiPlusFlat.h: No such file or directory
 #include <GdiPlusFlat.h>
                         ^
compilation terminated.
make[3]: *** [testfont] Error 1
```

hughbe:
It could be that I'm missing the file in the list [here](https://github.com/mono/libgdiplus/pull/54/files#diff-1a207f98e4c45ed0dbb9afa624e60016R20)

I pushed a change doing that. I'm trying to run the tests locally on WSL, but don't see any commands documented anywhere to test.

Do you know how I can run the tests? I'm probably doing something wrong

akoeplinger:
@hughbe `make check` should run these tests. Let's see what CI says :)

hughbe:
Looks like that made CI happy

PR #56, opened 2017-06-28
Fix missing symbols in region on Windows

hughbe:
@qmfrederik


hughbe:
I can merge them, but I wanted to break it up into bitesized related chunks that are specific to each component

It's kinda effort to combine them, but I don't mind if that's what you'd like

PR #72, opened 2017-07-02
Add more graphics tests

hughbe:
About half of these fail with libgdiplus - I'll submit a PR fixing them after this


akoeplinger:
Awesome, thanks so much for these ❤️

PR #75, opened 2017-08-10
Update docs for building on OSX/Linux

hughbe:
The old instructions were incorrect

I also expanded on the dependencies on OSX


akoeplinger:
Thanks!

PR #79, opened 2017-09-14
Fix some Clang warnings and static analyzer warnings

hughbe:
Unused variables and functions

### DIFF
--- a/src/bmpcodec.c
+++ b/src/bmpcodec.c
@@ -820,14 +820,20 @@ gdip_read_bmp_image (void *pointer, GpImage **image, ImageSource source)
 			// next three DWORD are the R,G,B masks, like bmiColors in BITMAPINFO
 			int size = sizeof (RGBQUAD);
 			size_read = gdip_read_bmp_data (pointer, (void*)&red_mask, size, source);
-			if (size_read != size)
+			if (size_read != size) {
+				status = InvalidParameter;
 				goto error;
+			}
 			size_read = gdip_read_bmp_data (pointer, (void*)&green_mask, size, source);
-			if (size_read != size)
+			if (size_read != size) {
+				status = InvalidParameter;
 				goto error;
+			}
 			size_read = gdip_read_bmp_data (pointer, (void*)&blue_mask, size, source);
-			if (size_read != size)
+			if (size_read != size) {
+				status = InvalidParameter;
 				goto error;
+			}
 		}
 
 		if ((red_mask == 0x7C00) && (green_mask == 0x3E0) && (blue_mask == 0x1F)) {
@@ -879,8 +885,10 @@ gdip_read_bmp_image (void *pointer, GpImage **image, ImageSource source)
 		/* stride is a (signed) _int_ and once multiplied by 4 it should hold a value that can be allocated by GdipAlloc
 		 * this effectively limits 'width' to 536870911 pixels */
 		size *= 4;
-		if (size > G_MAXINT32)
+		if (size > G_MAXINT32) {
+			status = InvalidParameter;
 			goto error;
+		}
 		result->active_bitmap->stride = size;
 		break;
 	}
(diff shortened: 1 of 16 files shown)


src/bmpcodec.c
```diff
@@ -899,7 +897,6 @@ gdip_read_bmp_image (void *pointer, GpImage **image, ImageSource source)
 
 		result->active_bitmap->palette = GdipAlloc (sizeof(ColorPalette) + sizeof(ARGB) * palette_entries);
 		if (result->active_bitmap->palette == NULL) {
-			status = OutOfMemory;
```
akoeplinger: instead of removing `status` in this function I think we should return it in the `error:` label below instead of the hardcoded `InvalidParameter`. At least that seems to me what the original intention was and also what the other functions do :)

src/bmpcodec.c
```diff
@@ -814,7 +813,6 @@ gdip_read_bmp_image (void *pointer, GpImage **image, ImageSource source)
 			/* the new structure contains the ARGB masks */
 			void *p = &bmi;
 			BITMAPV4HEADER *v4 = p;
-			alpha_mask = v4->bV4AlphaMask;
```
akoeplinger: I'd prefer keeping this since the comment above also mentions the `ARGB masks` and it'd be a bit strange to not read the alpha value then. Which warning complains about this?

src/bmpcodec.c
```diff
@@ -1224,8 +1215,10 @@ gdip_save_bmp_image_to_file_stream (void *pointer, GpImage *image, BOOL useFile)
 	gdip_bitmap_fill_info_header (image, &bmi);
 	gdip_write_bmp_data (pointer, (BYTE*) &bmi, sizeof (bmi), useFile);
 
-	if (colours) {
+    if (colours) {
```
akoeplinger: should be tabs instead of spaces

src/emfcodec.c
```diff
@@ -266,7 +266,6 @@ PolyPolygon (MetafilePlayContext *context, BYTE *data, BOOL compact)
 	n++;
 
 	/* total number of points (in all polygons)*/
-	int total = GETDW(DWP(n));
```
akoeplinger: this might be helpful for debugging and the comment above would be confusing then. Please just comment it out instead.

src/wmfcodec.c
```diff
@@ -59,6 +59,7 @@ gdip_getcodecinfo_wmf ()
 	return &wmf_codec;
 }
 
+#if DEBUG_WMF
```
akoeplinger: could you please move `#define GetParam(x,y) ...` from below into this ifdef as well since it's only used in DEBUG_WMF codepaths?

src/bmpcodec.c
hughbe: The warning is that `alpha_mask` was assigned but never used. The comment is still accurate though - we just don't use the A value

src/bmpcodec.c
hughbe: that makes sense, fixed

src/bmpcodec.c
hughbe: fixed

src/emfcodec.c
hughbe: fixed

src/wmfcodec.c
hughbe: Sure, fixed

src/bmpcodec.c
```diff
@@ -939,12 +939,12 @@ gdip_read_bmp_image (void *pointer, GpImage **image, ImageSource source)
...
 	size *= result->active_bitmap->height;
 	if (size > G_MAXINT32) {
-		status = OutOfMemory;		
+		status = OutOfMemory;
 		goto error;
 	}
 	pixels = GdipAlloc (size);
 	if (pixels == NULL) {
-		status = OutOfMemory;		
+		status = OutOfMemory;
```
hughbe: oops. Thanks!!

PR #84, opened 2017-09-20
[RFC] Add API for getting libgdiplus version

akoeplinger:
/cc @qmfrederik for feedback

This would simply return the libgdiplus version (e.g. `5.6`) as a string. It could be used like this:

```csharp

[DllImport("gdiplus")]
extern static string GetLibgdiplusVersion ();

Version LibgdiplusVersion {
    get {
        try {
            return new Version (GetLibgdiplusVersion ());
        } catch (EntryPointNotFoundException) {
            return null;  // in case of older libgdiplus or Windows GDI+
       }
   }
}
```

not sure if the string version is the best option or we should return major+minor as an int? or some other counter?

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -7,4 +7,4 @@ pkgconfig_DATA= libgdiplus.pc
 
 DISTCLEANFILES= libgdiplus.pc
 
-EXTRA_DIST = libgdiplus.pc.in LICENSE
+EXTRA_DIST = libgdiplus.pc.in LICENSE libgdiplus.sln winconfig.h.in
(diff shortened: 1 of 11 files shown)


qmfrederik:
@akoeplinger LGTM. Returing a string seems to be the most straightforward solution, let's keep it simple :)

qmfrederik:
As a side note - now that libgdiplus is living its renaissance :), it would make sense to get the distros to package updated versions of libgdiplus. Is there any kind of process to get that kicked off, or do you just need to contact the individual packages/maintainers directly?

akoeplinger:
> Is there any kind of process to get that kicked off, or do you just need to contact the individual packages/maintainers directly?

@qmfrederik yeah, though one of our engineers is the packager for most of these distros ;)

PR #28, opened 2014-10-30
Various fixes

PreferLinux:
Sorry that it is just two big commits...

Changes released under the MIT license.

PR #88, opened 2017-09-21
Fix invalid check in gdip_metafile_Arc

hughbe:
Extracted from https://github.com/mono/libgdiplus/pull/28


akoeplinger:
@hughbe re #28, would you be interested in taking a pass over it and extracting more valuable pieces?

PR #89, opened 2017-09-21
Add some more Pen tests and fix compat bugs

hughbe:


src/pen.c
```diff
@@ -253,6 +253,7 @@ GdipCreatePen1 (ARGB argb, REAL width, GpUnit unit, GpPen **pen)
 	result->color = argb;
 	/* FIXME: do unit conversion when setting width */
```
hughbe: I think this FIXME may be bogus. This is because the value of GdipGetPenWidth is the same for all unit values - this matches the Windows behaviour

@akoeplinger what do you think? do you know what unit conversion this could be referring to

akoeplinger: no idea but given this has been here for 14 years (https://github.com/hughbe/libgdiplus/commit/f20d007a91bd379991c3c2eb1f92bc0c5540f460) I'd say it's safe to assume nobody really needed a fix :P Let's get rid of it!

PR #90, opened 2017-09-22
Cleanup warnings

akoeplinger:
With these changes the build is now warning clean on OSX/clang for me.

### DIFF
--- a/src/font.c
+++ b/src/font.c
@@ -36,7 +36,11 @@
 #include "graphics-private.h"
 
 /* Generic fonts families */
+#if GLIB_CHECK_VERSION(2,32,0)
+static GMutex generic;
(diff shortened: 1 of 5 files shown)


src/font.c
```diff
@@ -195,8 +199,12 @@ GdipDeleteFontFamily (GpFontFamily *fontFamily)
 	if (!fontFamily)
 		return InvalidParameter;
 
+#if GLIB_CHECK_VERSION(2,32,0)
```
hughbe: I've seen this pattern all across the file. Maybe it's worth creating our own macro/function for this to avoid all these checks?

akoeplinger: yeah, makes sense.

PR #94, opened 2017-09-22
[Experiment/WIP] redefine wchar to be wchar_t

hughbe:
I realised that you can’t compile the same code with libgdiplus than with GDI+ if it uses WCHAR.

The problem is that WCHAR is defined as `wchar_t` on Windows GDI+, but is defined as `gunichar` in libgdiplus.

These types are not compatible (I think). They are certainly not source compatible, but I'm not sure about binary compatible (or whether that even matters).

This means that the following code compiles against GDI+, but fails with libgdiplus:

```cpp
int
main (int argc, char**argv)
{
  GdiplusStartupInput gdiplusStartupInput;
  ULONG_PTR gdiplusToken;
  GdiplusStartup (&gdiplusToken, &gdiplusStartupInput, NULL);

  GpImage *image;
  GdipLoadImageFromFile(L"test.ttf", &image);

  GdiplusShutdown (gdiplusToken);
  return 0;
}
```

I’m marking this PR as an experiment/WIP as I’m not sure the wider impact of this and wanted to get some eyes on this.

I’m also not sure about whether there is a combat issue from changing `gunichar` -> `wchar_t`.

The aim of the PR is to match Windows behaviour. I think this is important for libgdiplus. I personally think that source compatibility between libgdiplus and GDI+ could be important, but I’m not sure whether it is worth the risk.

This could also get rid of the external-facing dependency on glib.

### DIFF
--- a/src/font.c
+++ b/src/font.c
@@ -157,7 +157,7 @@ GdipPrivateAddFontFile (GpFontCollection *font_collection, GDIPCONST WCHAR *file
 	if (!font_collection || !filename)
 		return InvalidParameter;
     
-	file = (BYTE*) ucs2_to_utf8 ((const gunichar2 *)filename, -1);
+	file = (BYTE*) wchar_to_char (filename, -1);
 	if (!file)
 		return OutOfMemory;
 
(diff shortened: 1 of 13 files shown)


akoeplinger:
Looks like it breaks the tests :)

From what I read `wchar_t` is compiler dependent so I it's not guaranteed to be 16bits like guint16 what we're using now. I have to admit I'm not an expert in this area so not sure what the best solution is.

hughbe:
I'm closing this issue as I can't seem to figure out why the tests are failing.

PR #95, opened 2017-09-22
Enable graphics tests by fixing compatibility issues

hughbe:


src/graphics.c
```diff
@@ -2342,9 +2445,12 @@ GdipGetNearestColor (GpGraphics *graphics, ARGB *argb)
 GpStatus WINGDIPAPI
 GdipSetPageScale (GpGraphics *graphics, REAL scale)
 {
-	if (!graphics) 
+	if (!graphics || scale <= 0.0 || scale > 1000000032)
```
hughbe: I do not know where this number comes from... GDI+ accepts `1000000032` but rejects anything above it...

src/graphics.c
```diff
@@ -422,26 +432,37 @@ GdipDeleteGraphics (GpGraphics *graphics)
 		graphics->saved_status = NULL;
 	}	
 
+	graphics->state = GraphicsStateDeleted;
```
akoeplinger: @hughbe sorry for not catching this earlier, but this doesn't make sense. Since we're freeing `graphics` below you can't rely on the `graphics->state` being there after the call to `GdipDeleteGraphics()`.

tests/testgraphics.c
```diff
@@ -179,48 +180,39 @@ static void test_hdc ()
...
-// Libgdiplus does not validate that the graphics has an hdc before releasing.
-#if 0
 	status = GdipReleaseDC (graphics, hdc);
 	assert (status == InvalidParameter);	
-#endif
 
 	GdipDisposeImage (image);
 	GdipDeleteGraphics (graphics);
+
+	status = GdipGetDC (graphics, &hdc);
```
akoeplinger: This is one example, it's just pure luck that this still works in CI (it started randomly failing on my machine) since the memory could be overwritten at any time.

src/graphics.c
hughbe: That's ok. It's quite odd, because these methods always returns InvalidParameter with GDI+ when it is called on a deleted graphics.
This is probably undefined behaviour, right? What is the solution? Deleting these tests?

akoeplinger: Yes, this is undefined behavior since you have a dangling pointer after you've free'd the memory. I'd delete those tests and the GraphicsStateDeleted.

If it appears to work on GDI+ then it's probably just a coincidence as well. You can try allocating some chunks of memory after the free to see what happens.

hughbe: I'm just taking a look at the [Wine source code](https://github.com/wine-mirror/wine/blob/917e86dd7a4cca1da47262caaded2bef9e022f99/dlls/gdiplus/graphics.c#L2426-L2429)

```cpp
     /* Native returns ObjectBusy on the second free, instead of crashing as we'd
     * do otherwise, but we can't have that in the test suite because it means
     * accessing freed memory. */
    graphics->busy = TRUE;

    heap_free(graphics);

    return Ok;
```

This code is buggy (probably due to the fact that they can't test it...), as they use ObjectBusy, instead of InvalidParameter, but the point remains.

Maybe the right thing to do would be to keep these checks, but just remove the tests. This would be similar to what the wine folks did with their GDI+ port

akoeplinger: I still don't see the point, as soon as you free the pointer all bets are off. It looks like this was added to wine as part of a workaround for a buggy application that called GdipDeleteGraphics twice: https://bugs.winehq.org/show_bug.cgi?id=35006.

I don't think we should replicate this buggy behavior.

PR #96, opened 2017-09-22
Add --enable-warnaserror flag to configure and enable in CI

akoeplinger:
Ensures we build without warnings.


akoeplinger:

PR #97, opened 2017-09-25
Add Font tests, implement missing APIs, fix compat

hughbe:


src/font.c
```diff
@@ -1114,6 +1120,13 @@ gdip_create_font_from_logfont (void *hdc, void *lf, GpFont **font, BOOL ucs2)
 		result->face[LF_FACESIZE - 1] = '\0';
 	}
 
+#ifdef USE_PANGO_RENDERING
+	result->pango = NULL;
+#else
+	result->cairofnt = NULL;
+	result->cairo = NULL;
+#endif
+
```
hughbe: Without this change, we actually hit an assertion when deleting the font from inside cairo

tests/testfont.c
```diff
@@ -344,6 +354,465 @@ static void test_newInstalledFontCollection ()
...
+
+	GdipGetFontStyle (font, &style);
+	assert (style == 15);
+
+	GdipGetFontUnit(font, &unit);
+	assert (unit == UnitWorld);
+
+	status = GdipGetFamily (font, &family);
+	// FIXME: this fails with libgdiplus.
+#if defined(USE_WINDOWS_LIBGDIPLUS)
```
hughbe: Obvious bug here in libgdiplus - will fix after tho

tests/testfont.c
```diff
@@ -344,6 +354,465 @@ static void test_newInstalledFontCollection ()
...
```
hughbe: Same bug

tests/testfont.c
```diff
@@ -344,6 +354,465 @@ static void test_newInstalledFontCollection ()
...
+	GpFontFamily *originalFamily;
+	GpFont *font;
+	REAL height;
+
+	GdipGetGenericFontFamilySansSerif (&originalFamily);
+	GdipCreateFont (originalFamily, 10, 10, UnitPixel, &font);
+
+	status = GdipGetFontHeight (font, NULL, &height);
+	// FIXME: this returns InvalidParameter with libgdiplus.
+#if defined(USE_WINDOWS_LIBGDIPLUS)
```
hughbe: Another bug in libgdiplus, don't know how to fix

tests/testfont.c
```diff
@@ -344,6 +354,465 @@ static void test_newInstalledFontCollection ()
...
+	GpFontFamily *originalFamily;
+	GpFont *font;
+	REAL height;
+
+	GdipGetGenericFontFamilySansSerif (&originalFamily);
+	GdipCreateFont (originalFamily, 10, 10, UnitPixel, &font);
+
+	status = GdipGetFontHeightGivenDPI (font, 10, &height);
+	// FIXME: this returns InvalidParameter with libgdiplus.
+#if defined(USE_WINDOWS_LIBGDIPLUS)
```
hughbe: Same bug in libgdiplus, unsure how to fix

src/font.c
```diff
@@ -1083,7 +1089,7 @@ gdip_create_font_from_logfont (void *hdc, void *lf, GpFont **font, BOOL ucs2)
 	result->family = NULL;
 	/* Fixme - this is wrong, but I don't know of a quick way to get the emSize */
 	result->emSize = result->sizeInPixels;
-	result->unit = UnitPixel;
+	result->unit = UnitWorld;
```
akoeplinger: what's the reason for this change?

hughbe:
Yeah the tests are all the same thing - I think I've broken something in GdipGetFontHeightGivenDpi

src/font.c
hughbe: This is the default unit for a font created from a LOGFONT on windows

akoeplinger: ok, please mention that in the commit message.

hughbe:
I've just updated the PR to fix the unit test failures.

The problem is actually in existing libgdiplus code that indirectly calls GdipCreateFont. I put in a FIXME for later

PR #99, opened 2017-09-25
Add Matrix tests and fix compat bugs

hughbe:
Related: https://github.com/dotnet/corefx/pull/24238

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -17,7 +17,7 @@ LDADDS =					\
 	-lm
 
 noinst_PROGRAMS =			\
-	testbits testbrush testclip testfont testgraphics testhatchbrush testlineargradientbrush testpathgradientbrush testpen testpng testreversepath testsolidbrush testtexturebrush
+	testbits testbrush testclip testfont testgraphics testhatchbrush testlineargradientbrush testmatrix testpathgradientbrush testpen testpng testreversepath testsolidbrush testtexturebrush
 
 if HAS_X11
 noinst_PROGRAMS += testgdi
@@ -71,6 +71,12 @@ testlineargradientbrush_SOURCES =		\
 testlineargradientbrush_DEPENDENCIES = $(TEST_DEPS)
 testlineargradientbrush_LDADD = $(LDADDS)
 
+testmatrix_SOURCES =		\
+	testmatrix.c
+
+testmatrix_DEPENDENCIES = $(TEST_DEPS)
+testmatrix_LDADD = $(LDADDS)
+
 testreversepath_SOURCES =	\
 	testreversepath.c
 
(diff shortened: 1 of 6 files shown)


tests/testmatrix.c
```diff
@@ -0,0 +1,799 @@
...
+	assert (points[1].Y == 28);
+	assert (points[2].X == 28);
+	assert (points[2].Y == 40);
+
+	// Negative tests.
+	status = GdipTransformMatrixPointsI (NULL, points, 3);
+	assert (status == InvalidParameter);
+
+	// This causes a null pointer dereference in GDI+.
+#if !defined(USE_WINDOWS_LIBGDIPLUS)
```
hughbe: Another GDI+ bug that I don't think is worth replicating

tests/testmatrix.c
```diff
@@ -679,6 +679,7 @@ static void test_isMatrixIdentity ()
 	assert (isIdentity == TRUE);
 
 	// Less close to identity.
+	// FIXME: it appears that GDI+ is less precise than GDI+.
```
hughbe: @mellinoe @akoeplinger do you know the algorithm GDI+ uses for `GdipIsMatrixIdentity`? I've tried narrowing the values down to around `+-0.99949f` but I don't want to start mucking around with things that potentially break

mellinoe: Sorry, I don't know -- I'm really not familiar with the GDI+ codebase.

hughbe: No problem, just checking!

tests/testmatrix.c
```diff
@@ -0,0 +1,799 @@
...
+#include <math.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+int floatsEqual(float v1, float v2)
+{
+		return fabs (v1 - v2) < 0.0001;
+}
+
+void verifyMatrix (GpMatrix *matrix, REAL e1, REAL e2, REAL e3, REAL e4, REAL e5, REAL e6)
```
akoeplinger: this and floatsEqual already exist in testhelpers.h, please just #include that and use it from there (and while you're at it please get rid of the copy in testpathgradientbrush.c as well).

tests/Makefile.am
```diff
@@ -110,12 +116,13 @@ testpng_SOURCES =		\
 EXTRA_DIST =			\
 	$(testgdi_SOURCES)	\
 	$(testbits_SOURCES)	\
-	$(testbrush_SOURCES)	\
+	$(testbrush_SOURCES) \
```
akoeplinger: nit: revert the whitespace changes here and below.

hughbe:
Good point - I usually copy-paste the code into rextester.com to test the GDI+ code against Windows (then also on my local machine at the end to make sure). Rextester.com doesn't allow including files so I just copy-pasted them and forgot to revert that.

I've addressed the PR feedback, thanks for reviewing!

PR #100, opened 2017-09-25
Add AdjustableArrowCap tests and fix compat bugs

hughbe:
Builds on https://github.com/mono/libgdiplus/pull/99 to avoid merge conflicts

### DIFF
--- a/src/adjustablearrowcap.c
+++ b/src/adjustablearrowcap.c
@@ -169,7 +169,7 @@ GdipCreateAdjustableArrowCap (REAL height, REAL width, BOOL isFilled, GpAdjustab
 	cap->width = width;
 	cap->height = height;
 
-	if (height == 0)
+	if (width == 0)
 		cap->base.base_inset = 0;
 	else
 		cap->base.base_inset = height / width;
--- a/src/adjustablearrowcap.c
+++ b/src/adjustablearrowcap.c
@@ -169,6 +169,11 @@ GdipCreateAdjustableArrowCap (REAL height, REAL width, BOOL isFilled, GpAdjustab
 	cap->width = width;
 	cap->height = height;
 
+	if (height == 0)
+		cap->base.base_inset = 0;
+	else
+		cap->base.base_inset = height / width;
+
 	*arrowCap = cap;
 
 	return Ok;
--- a/src/customlinecap.c
+++ b/src/customlinecap.c
@@ -49,13 +49,12 @@ void
 gdip_custom_linecap_init (GpCustomLineCap *cap, CapClass *vt)
 {
 	cap->vtable = vt;
-	cap->base_cap = LineCapFlat;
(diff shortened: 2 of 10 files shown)


src/customlinecap.c
```diff
@@ -49,13 +49,12 @@ void
...
 	cap->vtable = vt;
-	cap->base_cap = LineCapFlat;
+	cap->base_cap = LineCapTriangle;
 	cap->start_cap = LineCapFlat;
 	cap->end_cap = LineCapFlat;
 	cap->stroke_join = LineJoinMiter;
 	cap->base_inset = 0.0;
-	/* LAMESPEC: Default value is documented as 1.0, but actually it is 0.0 */
-	cap->width_scale = 0.0;
+	cap->width_scale = 1.0;
```
hughbe: This and above now match GDI+ behaviour

src/customlinecap.h
```diff
@@ -14,6 +14,7 @@
 GpStatus WINGDIPAPI GdipCreateCustomLineCap (GpPath *fillPath, GpPath *strokePath, GpLineCap baseCap, float baseInset, GpCustomLineCap **customCap);
 GpStatus WINGDIPAPI GdipDeleteCustomLineCap (GpCustomLineCap *customCap);
 GpStatus WINGDIPAPI GdipCloneCustomLineCap (GpCustomLineCap *customCap, GpCustomLineCap **clonedCap);
+GpStatus WINGDIPAPI GdipGetCustomLineCapType( GpCustomLineCap* customCap, CustomLineCapType* capType);
```
akoeplinger: nit: space before the `(`, not after :)

akoeplinger: also `*` should be stuck to the argument, not the type, like everywhere else: `GpCustomLineCap *customCap`

src/customlinecap.c
```diff
@@ -349,6 +348,16 @@ GdipCloneCustomLineCap (GpCustomLineCap *customCap, GpCustomLineCap **clonedCap)
 }
 
 GpStatus WINGDIPAPI
+GdipGetCustomLineCapType (GpCustomLineCap* customCap, CustomLineCapType* capType)
```
akoeplinger: see comment about `*` placement below.

src/adjustablearrowcap.c
```diff
@@ -169,6 +169,11 @@ GdipCreateAdjustableArrowCap (REAL height, REAL width, BOOL isFilled, GpAdjustab
 	cap->width = width;
 	cap->height = height;
 
+	if (height == 0)
```
akoeplinger: shouldn't it check for width == 0 instead? otherwise we could have a divide by zero below.

hughbe: Thanks! This is totally a bug, my bad. I added some more edge-case tests that would have failed without this.

PR #102, opened 2017-09-26
Add FontFamily tests and fix source/behaviour compat bugs

hughbe:


src/font.c
```diff
@@ -453,6 +453,10 @@ gdip_font_clear_pattern_cache (void)
 static GpStatus
 create_fontfamily_from_collection (char* name, GpFontCollection *font_collection, GpFontFamily **fontFamily)
 {
+	/* Initialize the font set. */
+	if (font_collection->config)
+		gdip_createPrivateFontSet (font_collection);
```
akoeplinger: I wonder if we should move this up to `GdipCreateFontFamilyFromName()` instead? It'd be more consistent with  `GdipGetFontCollectionFamilyCount()` and `GdipGetFontCollectionFamilyList()`.

tests/testfont.c
```diff
@@ -823,6 +826,314 @@ static void test_getLogfontW ()
...
+	fontName = g_utf8_to_utf16 ("test.ttf", -1, NULL, NULL, NULL);
+	GdipPrivateAddFontFile (collection, fontName);
+
+	status = GdipCreateFontFamilyFromName (Tahoma, NULL, &family);
+	assert (status == Ok);
+	verifyFontFamily (family, "Tahoma", 2049, 423, 2472, "DejaVu Sans", 1901, 483, 2384);
+
+	status = GdipCreateFontFamilyFromName (CodeNewRoman, collection, &family);
+	assert (status == Ok);
+	verifyFontFamily (family, "Code New Roman", 2059, 430, 2489, "Code New Roman", 1901, 483, 2384);
```
filipnavara: @hughbe Where do these values come from? I have been doing some text with macOS+Pango (--with-pango option) and on Windows. In both cases the reported values for ascent and descent were 1884 and 514, which are specified in the font in the TT_OS2 table in usWinAscent and usWinDescent fields. Maybe I am missing something since the test obviously passes on some platforms with the cairo backend.

filipnavara: Oh, it actually fails only in the macOS + Pango + CoreText backend. Otherwise the metrics are calculated from FreeType in the same way as the default Cairo backend.

hughbe: The font tests do need a bit of work - most of the currently fail on Windows against libgdiplus.

As per where the values came from, I ran the tests against GDI+ and inputted the values returned from GDI+

filipnavara: That's the weird thing, because there are two sets of values. I expected one of the to match current libgdiplus cairo implementation and the other to match whatever GdiPlus.dll returns on Windows. That just doesn't happen on my machine though.

filipnavara: Turns out that the first set of values is actually for Tahoma. Both the default cairo backend and pango w/ FC backend incorrectly maps to Tahoma instead of Code New Roman.

hughbe: That's a pretty bad, right? CodeNewRoman should be fetched from the font collection... shame I didn't put more care into testing this

filipnavara: Yep, it's a bug. On Cairo backend it stems from the cairo_ft_font_face_create_for_pattern call, which ignore the private font collections.

PR #104, opened 2017-09-26
Add apt-get instructions in README

hughbe:


README.md
```diff
@@ -6,11 +6,16 @@ This is part of the [Mono project](http://www.mono-project.com/).
 
 This requires the libraries used by the Cairo vector graphics library to build (freetype2, fontconfig, Xft2 and libpng).
 
-On OSX you can use [Homebrew](https://brew.sh/) to install the dependencies:
+On **OSX** you can use [Homebrew](https://brew.sh/) to install the dependencies:
 
 	brew install glib cairo libexif libjpeg giflib libtiff autoconf libtool automake
 
-On Windows you can use [Chocolatey](https://chocolatey.org) to install the dependencies. Run the following commands from the root of the repository from an admin command prompt:
+On **Linux** you can use `apt-get` to install the dependencies:
```
akoeplinger: apt-get is only used on Debian-based distros so probably better to be explicit:

`On Debian-based Linux distributions you can use apt-get to install the dependencies`

README.md
```diff
@@ -6,11 +6,16 @@ This is part of the [Mono project](http://www.mono-project.com/).
...
+
+	sudo apt-get update
```
akoeplinger: I wouldn't mention this, we can assume the apt cache is up to date on a normal developer machine.

README.md
```diff
@@ -6,11 +6,16 @@ This is part of the [Mono project](http://www.mono-project.com/).
...
+	sudo apt-get install libgif-dev xvfb autoconf libtool automake build-essential gettext cmake git libglib2.0-dev libcairo2-dev libtiff-dev libexif-dev
```
akoeplinger: xvfb shouldn't be needed, I guess this just brought in some necessary dependency?

Same for cmake and git.

hughbe: Only included these as they were included in the `.travis.yml` file:
https://github.com/mono/libgdiplus/blob/master/.travis.yml

README.md
hughbe: Removed

README.md
hughbe: Right - fixed

README.md
akoeplinger: Ah. Yeah those are just there so we can compile Mono and run the System.Drawing+WinForms tests :)

PR #105, opened 2017-09-27
Define custom assertions with better error messages

hughbe:
The first commit deals with test compilation failures running the tests against GDI+.

The main bulk of this PR is in the 3rd commit, which defines better error messages instead of the default message that doesn’t show much information.

### DIFF
--- a/tests/testhelpers.h
+++ b/tests/testhelpers.h
@@ -2,15 +2,15 @@
 #include <float.h>
 #include <math.h>
 
-int floatsEqual(float v1, float v2)
+BOOL floatsEqual (float v1, float v2)
 {
-    if (isnan (v1))
-        return isnan (v2);
+	if (isnan (v1))
+		return isnan (v2);
 
-    if (isinf (v1))
-        return isinf (v2);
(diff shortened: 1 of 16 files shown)


tests/testhelpers.h
```diff
@@ -40,3 +40,29 @@ void verifyMatrix (GpMatrix *matrix, REAL e1, REAL e2, REAL e3, REAL e4, REAL e5
...
 #endif
+
+#define assertEqualInt(actual, expected) assertEqualIntImpl (actual, expected, __FILE__, __LINE__)
+void assertEqualIntImpl (INT actual, INT expected, const char *file, INT line)
+{
+    if (actual != expected)
+    {
+        printf ("Assertion failed on line %d in %s\n", line, file);
+        printf ("Expected: %d\n", actual);
+        printf ("Actual:   %d\n", expected);
```
akoeplinger: - You switched actual and expected here.
- We should write to stderr instead of stdout like `assert()`

hughbe: Thanks for noticing. I've fixed this.

PR #106, opened 2017-09-27
Fix remaining Clang warnings

hughbe:


src/jpegcodec.c
```diff
@@ -659,8 +654,7 @@ gdip_save_jpeg_image_internal (FILE *fp, PutBytesDelegate putBytesFunc, GpImage
 	struct jpeg_compress_struct	cinfo;
 	struct gdip_jpeg_error_mgr	jerr;
 	const EncoderParameter		*param;
-	GpBitmap	*bitmap = (GpBitmap *) image;
-	JOCTET		*scanline = NULL;
+    JOCTET		*scanline = NULL;
```
akoeplinger: indentation

src/region-bitmap.c
```diff
@@ -529,7 +529,7 @@ gdip_region_bitmap_from_path (GpPath *path)
 
 	/* here the memory is allocated for a ARGB bitmap - so 32 times bigger than our normal alpha bitmap */
 	if ((size < 1) || (size > REGION_MAX_BITMAP_SIZE << 5)) {
-		g_warning ("Path conversion requested %d bytes (%d x %d). Maximum size is %d bytes.",
+        g_warning ("Path conversion requested %lu bytes (%d x %d). Maximum size is %d bytes.",
```
akoeplinger: indentation

hughbe:
Fixed the indentation. I was experimenting with a [CMake build system](https://github.com/hughbe/libgdiplus/tree/cmake-experiment) (I don't think a PR would be accepted to actually replace the Automake system with a CMake build system but I wanted to create an XCode project so I could debug the code more easily).

For some reason, these warnings popped up in the XCode project. To that end, I'm using `AppleClang 9.0.0.9000037 / OSX 10.12.6 (16G29) `

akoeplinger:
> I don't think a PR would be accepted to actually replace the Automake system with a CMake build system

Yeah not at this time. We've been thinking about moving to CMake in Mono for some time but it always gets put on the back burner. If/should that happen I guess we can move libgdiplus over as well :)

Thanks for the link, I've tried adding [`-pedantic`](https://github.com/hughbe/libgdiplus/blob/cmake-experiment/CMakeLists.txt#L79) and that showed _some_ of the warnings but not all.

I assume CMake sets some other flags by default, could you please check the build log and see what flags are actually passed to clang?

PR #107, opened 2017-09-27
Add CustomLinecap tests

hughbe:


tests/testcustomlinecap.c
```diff
@@ -0,0 +1,571 @@
...
+	LineCap baseCap;
+	REAL baseInset;
+	LineCap startCap;
+	LineCap endCap;
+	LineJoin strokeJoin;
+	REAL widthScale;
+	CustomLineCapType type;
+
+	// GDI+ returns 0 here each time. This looks like a GDI+ bug.
+#if !defined(_WIN32)
```
akoeplinger: should probably be `!defined(USE_WINDOWS_LIBGDIPLUS)` right? I see a few more of these, we can do a separate cleanup pass.

PR #109, opened 2017-09-28
Enable extra warnings

hughbe:
These I believe are set to the default with CMake (maybe not pedantic).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -31,7 +31,7 @@ PKG_CHECK_MODULES(CAIRO, cairo >= $CAIRO_REQUIRED_VERSION)
 # Optional use (experimental and unsupported) of Pango's text rendering on top of Cairo
 AC_ARG_WITH(pango, [  --with-pango],[text_v=pango],[text_v=cairo])
 
-GDIPLUS_CFLAGS="$GDIPLUS_CFLAGS -Wall -Wextra -Wno-unused-parameter -Wno-sign-compare"
+GDIPLUS_CFLAGS="$GDIPLUS_CFLAGS -Wall -Wextra -Wno-unused-parameter -Wno-sign-compare -std=gnu99"
 
 AC_ARG_ENABLE(warnaserror, AS_HELP_STRING([--enable-warnaserror],[Enable treating warnings as errors.]),[warnaserror=yes],[warnaserror=no])
 
--- a/src/carbon-private.c
+++ b/src/carbon-private.c
@@ -40,4 +40,10 @@ void gdip_get_display_dpi_carbon (float *h_dpi, float *v_dpi) {
 		}
 	}
 }
+#else
+#include <glib.h>
+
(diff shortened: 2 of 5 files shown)


src/Makefile.am
```diff
@@ -121,6 +121,6 @@ libgdiplus_la_SOURCES = \
 
 libgdiplus_la_LIBADD = $(GDIPLUS_LIBS)
 
-AM_CPPFLAGS = $(GDIPLUS_CFLAGS) -Wall -Wno-unused -Wno-format
+AM_CPPFLAGS = $(GDIPLUS_CFLAGS) -Wall -Wno-unused -Wno-format -pedantic
```
akoeplinger: this isn't necessary since it's already included in GDIPLUS_CFLAGS

configure.ac
```diff
@@ -29,7 +29,7 @@ PKG_CHECK_MODULES(CAIRO, cairo >= $CAIRO_REQUIRED_VERSION)
 # Optional use (experimental and unsupported) of Pango's text rendering on top of Cairo
 AC_ARG_WITH(pango, [  --with-pango],[text_v=pango],[text_v=cairo])
 
-GDIPLUS_CFLAGS="$GDIPLUS_CFLAGS -Wall -Wextra -Wno-unused-parameter -Wno-sign-compare"
+GDIPLUS_CFLAGS="$GDIPLUS_CFLAGS -Wall -Wextra -Wno-unused-parameter -Wno-sign-compare -pedantic"
```
akoeplinger: according to the Linux build failure we need to add `-std=c99` so it doesn't default to c89 and complains about ancient things.

akoeplinger:
@hughbe ok so I looked at the Linux build and there are two classes of warnings left that get triggered by pedantic:

```
bmpcodec.c:1186:4: error: ISO C forbids conversion of object pointer to function pointer type [-Werror=pedantic]
00:21:03    ((PutBytesDelegate)(pointer))(data, size); 
00:21:03     ^
00:21:03 bmpcodec.c: In function 'gdip_save_bmp_image_to_stream_delegate':
```

I can't wrap my head around these at the moment, any ideas?

The other one are warnings like:

```
testpen.c: In function 'test_setPenMode':
testpen.c:1301:2: error: overflow in implicit constant conversion [-Werror=overflow]
  assertEqualInt (mode, (PenAlignment)(PenAlignmentCenter - 1));
  ^
```

That one makes more sense to me and is fixed by removing the cast to `PenAlignment`.

Do you think we should look into this further or is it "too pedantic" at this point? 😄 

If I fix the overflow warnings and uncomment the locations where the *BytesDelegate are used then it compiles to the end at least.

hughbe:
I agree. Pedantic is certainly unecessary for the unit tests. For the product code it may be good. But it looks unecessary. I reverted this change but fixed some of the warnings pedantic wuld have found.

PR #110, opened 2017-09-28
Add StringFormat tests and compat bugs

hughbe:


tests/teststringformat.c
```diff
@@ -0,0 +1,742 @@
...
+	if (isnan (v1))
+		return isnan (v2);
+
+	if (isinf (v1))
+		return isinf (v2);
+
+	return fabs (v1 - v2) < 0.0001;
+}
+
+void verifyMatrix (GpMatrix *matrix, REAL e1, REAL e2, REAL e3, REAL e4, REAL e5, REAL e6)
```
akoeplinger: use testhelper.h ?

tests/teststringformat.c
```diff
@@ -0,0 +1,742 @@
...
+        fprintf (stderr, "Actual:   %d\n", actual);
+
+#if !defined(_WIN32)
+				abort ();
+#endif
+    }
+}
+
+#define assertEqualFloat(actual, expected) assertEqualFloatImpl (actual, expected, __FILE__, __LINE__)
+void assertEqualFloatImpl (REAL actual, REAL expected, const char *file, INT line)
```
akoeplinger: same here.

src/stringformat-private.h
```diff
@@ -44,6 +44,7 @@ struct _StringFormat {
 	StringFormatFlags	formatFlags;
 	StringTrimming		trimming;
 	StringDigitSubstitute	substitute;
+	LANGID		language;
```
akoeplinger: please fix indendation so it lines up with the others

hughbe: I am the worst at indentation, sorry! For some reason VS code sometimes gets it right but often doesn't. It also doesn't seem to be able to line up tabs correctly

akoeplinger: yeah I see the same in VS Code. nevermind ❤️

PR #112, opened 2017-09-30
Add more Image tests and fix some compat bugs

hughbe:
A bunch of bugs here that I don’t know how to fix though


hughbe:
Test failure
```
1) Icon (MonoTests.System.Drawing.GDIPlusTest.Icon)
     size
  Expected: 8
  But was:  12
  at MonoTests.System.Drawing.GDIPlusTest.Icon () [0x001db] in /home/travis/build/mono/mono/mcs/class/System.Drawing/Test/System.Drawing/GDIPlusTest.cs:874 
  at (wrapper managed-to-native) System.Reflection.MonoMethod.InternalInvoke(System.Reflection.MonoMethod,object,object[],System.Exception&)
  at System.Reflection.MonoMethod.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x00032] in /home/travis/build/mono/mono/mcs/class/corlib/System.Reflection/MonoMethod.cs:305 
```

Do these tests run against Windows GDI+? I tried running the following, but got the size of `12` against Windows GDI+ (the unit tests expect 8, but I fixed this to 12)

Maybe this is another case where we need updates on the Mono size

[ConsoleApp1.zip](https://github.com/mono/libgdiplus/files/1346716/ConsoleApp1.zip)

akoeplinger:
Yeah there appear to be multiple issues in this test when run on Windows, like the `Assert.AreEqual(1032, size)` which are size=12 on Windows.

The `GDIPlus.GdipSetImagePalette(bitmap, palette)` before the `Assert.AreEqual(8, size)` actually fails as well with InvalidParameter status code.

src/image.c
```diff
@@ -1975,7 +1978,11 @@ GdipGetImagePaletteSize (GpImage *image, INT* size)
         if (image->active_bitmap->pixel_format == PixelFormat4bppIndexed)
                 palette_entries = 16;
 
-	*size = (palette_entries - 1) * sizeof(ARGB) + sizeof(ColorPalette);
+	if (palette_entries == 0)
+		*size = sizeof(ColorPalette);
+	else
+		*size = (palette_entries - 1) * sizeof(ARGB) + sizeof(ColorPalette);
+
```
hughbe: This is the "fix" that caused the tests to fail.

Previously, we'd actually subtract(!) from the total size of ColorPalette, leading to potentially under allocating memory.

akoeplinger: Yeah I agree this is a good fix, the question is just what the best way of coordinating the fix with updating the unit tests in Mono is.

akoeplinger: we do have a collection of FIXME's already that require changes to the Mono unit tests right? maybe we should batch them together and enable them all in one go with the next release?

akoeplinger: Hmm, actually if you look at the comment in GdipGetImagePalette it seems to explicitly mention and contradict the fix:

```
 		/* ColorPalette definition includes one ARGB member (ARGB Entries[1]). In reality there are 
 		 * Count entries allocated, where Count can be 0 (and required to substract the ARGB size)
 		 */
 		/* coverity[buffer_alloc] */
 		image->active_bitmap->palette = (ColorPalette*) GdipAlloc (sizeof(ColorPalette) - sizeof(ARGB));
```

It allocates `-1 * sizeof(ARGB) + sizeof(ColorPalette)` there!

akoeplinger:
So I ran the GDIPlusTest.cs on Windows and the following 11 out of 52 tests fail there: 

image

src/image.c
hughbe: Okay so I did a bit of digging in to this. The fact is that in the following code, `GdipGetImagePaletteSize` returns `sizeof(ColorPalette)` if there is the current count is 0.  [Wine also has this check](https://github.com/wine-mirror/wine/blob/c579a81a6353dd7c843270561cd5763ff895de40/dlls/gdiplus/image.c#L2291-L2294).

```
static void test_getImagePaletteSize ()
{
	GpStatus status;
	GpImage *bitmapImage = getImage ("test.bmp");
	INT size;

	status = GdipGetImagePaletteSize (bitmapImage, &size);
	assertEqualInt (status, Ok);
	assertEqualInt (sizeof(ColorPalette), 12);
	assertEqualInt (size, 12);

	GdipDisposeImage (bitmapImage);
}
```

However, I was pretty much wrong on all fronts in GdipGetImagePalette. I've added a bunch of tests in a new commit. All of these failed previously with my change, and also with my fix...

Take a look and let me know

hughbe:
Thanks for running those tests. I'm going to uni on Monday so I'm probably not going to have time to investigate them by the way! I'm just trying to send off PRs before then :D

akoeplinger:
@hughbe Understood, thank you so much for the work you've been doing so far and enjoy your time at uni 👍

src/bitmap.c
```diff
@@ -1024,21 +1024,10 @@ GdipCreateBitmapFromHICON (HICON hicon, GpBitmap** bitmap)
...
-		 */
-		/* coverity[buffer_alloc] */
-		ColorPalette *palette = (ColorPalette*)GdipAlloc (sizeof (ColorPalette) - sizeof (ARGB));
-		if (!palette)
-			return OutOfMemory;
-
-		palette->Flags = 0;
-		palette->Count = 0;
-		status = GdipSetImagePalette ((GpImage*)*bitmap, palette);
+		(*bitmap)->active_bitmap>palette = NULL;
```
akoeplinger: missing the dash in `->`.

also should we free the old `palette` before setting it to NULL?

src/image.c
```diff
@@ -1903,21 +1903,21 @@ GdipGetImagePalette (GpImage *image, ColorPalette *palette, INT size)
...
 
 	if (!image->active_bitmap->palette) {
-		/* ColorPalette definition includes one ARGB member (ARGB Entries[1]). In reality there are 
-		 * Count entries allocated, where Count can be 0 (and required to substract the ARGB size)
-		 */
-		/* coverity[buffer_alloc] */
-		image->active_bitmap->palette = (ColorPalette*) GdipAlloc (sizeof(ColorPalette) - sizeof(ARGB));
-		image->active_bitmap->palette->Flags = 0;
-		image->active_bitmap->palette->Count = 0;
+		if (size >= 0 && size < sizeof(ColorPalette))
```
akoeplinger: so passing `size` smaller than zero is an accepted input or do we need an additional check?

src/image.c
```diff
@@ -1940,7 +1939,7 @@ GdipSetImagePalette (GpImage *image, GDIPCONST ColorPalette *palette)
 {
 	int size;
 
-        if (!image || !palette)
+    if (!image || !palette || palette->Count == 0 || palette->Count > 256)
```
akoeplinger: indentation

tests/testgpimage.c
```diff
@@ -564,28 +592,28 @@ static void test_getEncoderParameterListSize ()
 	// FIXME: this returns NotImplemented with libgdiplus.
 #if defined(USE_WINDOWS_GDIPLUS)
 	assertEqualInt (status, Ok);
-	assertEqualInt (size, 164);
+	assertEqualInt (size, is_32bit() ? 164 : 184);
 #endif
 
 	status = GdipGetEncoderParameterListSize (image, &gifEncoderClsid, &size);
-// FIXME: this returns FileNotFound with libgdiplus.
+	// FIXME: this returns FileNotFound with libgdiplus.
```
akoeplinger: bunch of inadvertent indentation changes here and below.

src/image.c
hughbe: The following test passes against Windows GDI+. Negative sizes seem to be ignored.
```

	// Empty palette - negative size.
	palette->Count = 100;
	status = GdipGetImagePalette (bitmapImage, palette, -1);
	assertEqualInt (status, Ok);
	assertEqualInt (palette->Count, 0);
```

tests/testgpimage.c
hughbe: They were deliberate as this now aligns with the code below.

tests/testimageattributes
```diff
@@ -0,0 +1,210 @@
+#! /bin/sh
```
akoeplinger: this shouldn't be checked in

hughbe: Oops.. I've removed this. Good news is that now everything looks like it works :D

PR #113, opened 2017-09-30
Fix the width and height of EMF images

hughbe:
Now actually makes sense!


src/metafile.c
```diff
@@ -1343,19 +1343,21 @@ gdip_get_metafile_from (void *pointer, GpMetafile **metafile, ImageSource source
...
-	case METAFILETYPE_EMFPLUSDUAL:
+	case MetafileTypeEmf:
+	case MetafileTypeEmfPlusOnly:
+	case MetafileTypeEmfPlusDual:
 		mf->base.image_format = EMF;
 		mf->length = mf->metafile_header.Header.Emf.nBytes - mf->metafile_header.Header.Emf.nSize;
 		adjust_emf_headers = TRUE;
 		break;
+    default:
+        goto error;
```
akoeplinger: status needs to be set here otherwise it'll return Ok status from previous call to gdip_get_metafileheader_from

akoeplinger: indentation needs some love too 😄

src/gdipenums.h
```diff
@@ -736,4 +736,37 @@ typedef enum {
...
 
+typedef enum MetafileType
+{
+    MetafileTypeInvalid,
+    MetafileTypeWmf,
+    MetafileTypeWmfPlaceable,
+    MetafileTypeEmf,
+    MetafileTypeEmfPlusOnly,
+    MetafileTypeEmfPlusDual
+} MetafileType;
```
akoeplinger: the other enums are indented by tabs, please do that here as well

src/metafile.c
```diff
@@ -1275,18 +1275,12 @@ g_warning ("EMF HEADER iType %d, nSize %d, Bounds L %d, T %d, R %d, B %d, Frame
 			return InvalidParameter;
 
 		header->Type = MetafileTypeEmf;
-		/* FIXME: inclusive-inclusive, MS gets 1 to 5 pixels larger than the header */
-		if ((emf->rclFrame.left == 0) && (emf->rclFrame.top == 0)) {
-			header->X = 0;
-			header->Y = 0;
-			header->Width = emf->szlDevice.cx + 1;
-			header->Height = emf->szlDevice.cy + 1;
```
akoeplinger: are you sure removing this case is ok now? (quite late here already so apologies if this is obvious 😆 )

hughbe: This is the actual bug fix to the previously disabled test. If you look at the [MSDN docs](https://msdn.microsoft.com/en-us/library/windows/desktop/ms534065(v=vs.85).aspx)

> **szlDevice**
>     Type: **SIZEL**
>     Resolution, in pixels, of the reference device.

So basically the problem here was that the width and height of the EMF file would actually be the resolution of my screen (1440x900 because I created the file on my laptop).

This is the wrong parameter to use for width and height. I just combined these two cases into the same one, because the `x` and `y` value would be `0`, so subtracting`0` would not change the value.

PR #115, opened 2017-10-01
Add ImageAttribute tests and fix compat bugs

hughbe:
Most are negative tests here

### DIFF
--- a/src/imageattributes.c
+++ b/src/imageattributes.c
@@ -337,8 +337,50 @@ GdipDisposeImageAttributes (GpImageAttributes *imageattr)
 }
 
 GpStatus WINGDIPAPI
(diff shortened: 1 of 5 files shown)


src/imageattributes.c
```diff
@@ -337,8 +337,50 @@ GdipDisposeImageAttributes (GpImageAttributes *imageattr)
 }
 
 GpStatus WINGDIPAPI
+GdipSetImageAttributesToIdentity (GpImageAttributes *imageattr, ColorAdjustType type)
+{
+	GpImageAttribute *imgattr;
```
akoeplinger: OMG this variable name is so easy to confuse with the `imageattr` method argument. Not your fault of course since it's in the existing code too but maybe we should consider renaming these in the future :)

PR #116, opened 2017-10-01
Implement GdipCloneFont

hughbe:


hughbe:
Hmmm I can't quite understand the test failure...

```
lt-testfont: fccache.c:534: FcCacheFini: Assertion `fcCacheChains[i] == NULL' failed.
FAIL testfont (exit status: 134)
```

On Linux only, too. Hmm

src/font.c
```diff
@@ -927,11 +932,57 @@ GdipCreateFont (GDIPCONST GpFontFamily* family, float emSize, int style, Unit un
...
+	result->emSize = font->emSize;
+	result->unit = font->unit;
+
+	result->face = GdipAlloc (strlen ((char *) font->face) + 1);
+	if (!result->face) {
+		GdipFree (result);
+		return OutOfMemory;
+	}
+
+	memcpy (result->face, font->face, strlen ((char *) font->face) + 1);
```
akoeplinger: I think we could just use g_strdup instead of doing the memcopy ourselves.

src/font.c
```diff
@@ -932,6 +932,51 @@ GdipCreateFont (GDIPCONST GpFontFamily* family, float emSize, int style, Unit un
...
+
+	status = GdipCloneFontFamily (font->family, &result->family);
+	if (status != Ok) {
+		GdipFree (result);
+		GdipFree (result->face);
```
akoeplinger: this should be the other way around otherwise `result` will already have been freed.

tests/testfont.c
```diff
@@ -103,6 +103,8 @@ static void verifyFont (GpFont *font, GpFontFamily *originalFamily, INT expected
 	status = GdipGetFontStyle(font, &style);
 	assertEqualInt (status, Ok);
 	assertEqualInt (style, expectedStyle);
+	
+	GdipDeleteFontFamily (family);
```
akoeplinger: I looked at it with gdb and it was crashing in the `FcFini()` call in `GdiplusShutdown()`, presumably because this wasn't cleaned up properly.

PR #118, opened 2017-10-15
Add BMP codec tests and fix compat bugs

hughbe:


hughbe:
> /usr/local/Homebrew/Library/Homebrew/brew.rb:12:in `<main>': Homebrew must be run under Ruby 2.3! (RuntimeError)

Unrelated CI failure?

src/bmpcodec.c
```diff
@@ -118,9 +118,8 @@ gdip_get_bmp_pixelformat (BITMAPINFOHEADER *bih, PixelFormat *dest)
 		case 1:
 			*dest = PixelFormat1bppIndexed;
 			break;
-	        default:
-			g_warning ("Unsupported bitcount (%d) and/or compression (%d).", bitCount, compression);
```
akoeplinger: I'd keep the warning message.

tests/.gitignore
```diff
@@ -3,8 +3,10 @@
 .libs/
 Makefile
 Makefile.in
+temp.*
```
akoeplinger: Let's just delete temp.bmp at the end of the test instead.

src/bmpcodec.c
hughbe: This does end up causing messages to be printed to the log/console though. I can add back if you'd like but I'm not sure I see the value

tests/.gitignore
hughbe: I added some code to delete the file at the end of the test, but also kept this - if the tests fail, we abort so the assets may not be cleaned up

src/bmpcodec.c
akoeplinger: hmm I guess the value was in making sure all the possible combinations are covered. If we're sufficiently sure we covered everything then I'm fine with removing the warning.

tests/.gitignore
akoeplinger: I don't care about that since it's not the normal case, please remove it :)

tests/testbmpcodec.c
```diff
@@ -23,17 +23,22 @@ using namespace DllExports;
 #include <stdlib.h>
 #include "testhelpers.h"
 
-static WCHAR file[] = {'t', 'e', 'm', 'p', '.', 'b', 'm', 'p', 0};
+#if !defined(_WIN32)
+#include <unistd.h>
+#endif
+
+static const char *file = "temp/temp.bmp";
+static WCHAR wFile[] = {'t', 'e', 'm', 'p', '/', 't', 'e', 'm', 'p', '.', 'b', 'm', 'p', 0};
```
akoeplinger: the tests seems to fail, probably because you added it in a subfolder which fopen doesn't automatically create. I'd instead just keep it in the same folder and name it more uniquely, e.g. testbmpcodec.bmp

PR #120, opened 2017-10-25
Fix cloned metafile raw format

hughbe:
Previously this wasn’t being set when we cloned the metafile.

### DIFF
--- a/src/metafile.c
+++ b/src/metafile.c
@@ -873,6 +873,7 @@ gdip_metafile_clone (GpMetafile *metafile, GpMetafile **clonedmetafile)
 	if (!mf)
 		return OutOfMemory;
 
+	memcpy (&mf->base.image_format, &metafile->base.image_format, sizeof (GUID));
 	memcpy (&mf->metafile_header, &metafile->metafile_header, sizeof (MetafileHeader));
 	if (metafile->length > 0) {
 		mf->data = GdipAlloc (metafile->length);
--- a/src/metafile.c
+++ b/src/metafile.c
@@ -855,6 +855,8 @@ gdip_metafile_create ()
 {
 	GpMetafile* mf = (GpMetafile*) GdipAlloc (sizeof (GpMetafile));
 	if (mf) {
+		memset (&mf->base, 0, sizeof (GpImage));
+
 		mf->base.type = ImageTypeMetafile;
 		mf->delete = FALSE;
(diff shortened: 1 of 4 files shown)


src/metafile.c
```diff
@@ -873,9 +873,15 @@ gdip_metafile_clone (GpMetafile *metafile, GpMetafile **clonedmetafile)
 	if (!mf)
 		return OutOfMemory;
 
+	memcpy (&mf->base.image_format, &metafile->base.image_format, sizeof (GUID));
```
akoeplinger: why is copy base.image_format special, don't we need all other fields too?

hughbe: No other fields are used. Do you think we should just copy M.D.->image?

akoeplinger: I see e.g. num_of_frames used a couple times in the codebase, not sure why you think only image_format is used :)

I assume we need to do a deep copy of all the fields, otherwise it wouldn't be a true clone?

hughbe: > I see e.g. num_of_frames used a couple times in the codebase, not sure why you think only image_format is used :)

I believe that for each of these, this is hidden behind the `ImageTypeBitmap` case.

But maybe it's just a good idea to do deep copy everything

akoeplinger: Yeah. Let me know if you want to do this as part of this PR or separately, I'm fine with merging the current state since it's already better than what was there before :)

src/metafile.c
```diff
@@ -870,10 +872,13 @@ GpStatus
 gdip_metafile_clone (GpMetafile *metafile, GpMetafile **clonedmetafile)
 {
 	GpMetafile *mf = gdip_metafile_create ();
+	GpImage *base;
 	if (!mf)
 		return OutOfMemory;
 
-	memcpy (&mf->base.image_format, &metafile->base.image_format, sizeof (GUID));
+	gdip_bitmap_clone (&metafile->base, &base);
```
akoeplinger: do we need `gdip_bitmap_setactive(*cloneImage, NULL, 0);` like  in GdipCloneImage () as well, or should we just call that method instead?

we should probably also assert that `metafile->base->type == ImageTypeBitmap`.

hughbe: The base type actually isn’t Bitmap, it’s metafile, so this assertion would be invalid.

PR #121, opened 2017-10-26
Fix GdipGetImageDimension for metafiles to match GDI+

hughbe:
Relies on #120

Previously, we’d be off by a factor of about 0.02-0.05.

I looked at the docs and discovered that GdipGetImageDimension for metafiles basically returns the width and height in 0.01 millimetre units.

So to get the dimensions, we can just convert the pixels into millimetres using the DPI, and then multiply that by 100 to get the right units.

This is now exactly the value that GDI+ returns, so we don’t need to nerf the tests for this behaviour difference. I’ve updated the tests to reflect the exact values returned by GDI+, that are now also returns by libgdiplus

### DIFF
--- a/src/image.c
+++ b/src/image.c
@@ -1257,8 +1257,11 @@ GdipGetImageDimension (GpImage *image, REAL *width, REAL *height)
 		break;
 	case ImageTypeMetafile: {
 		MetafileHeader *metaheader = gdip_get_metaheader(image);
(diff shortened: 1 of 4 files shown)


akoeplinger:
FYI I'm archiving the tests/test-suite.log file on Jenkins now so you can actually take a look at which assertion failed: https://jenkins.mono-project.com/view/Pull%20requests/job/test-libgdiplus-pull-request/label=ubuntu-1404-amd64/230/artifact/tests/test-suite.log

PR #122, opened 2017-11-13
Fixup whitespace in source code

hughbe:
This makes the code a lot more readable and consistent - not sure if we want this in the git history but putting it out here anyway

### DIFF
--- a/src/bitmap.c
+++ b/src/bitmap.c
@@ -75,7 +75,7 @@ gdip_is_a_supported_pixelformat (PixelFormat fmt)
 	case PixelFormat1bppIndexed:
 	case PixelFormat4bppIndexed:
 	case PixelFormat8bppIndexed:
- 	case PixelFormat24bppRGB:
+	case PixelFormat24bppRGB:
 	case PixelFormat32bppARGB:
 	case PixelFormat32bppPARGB:
 	case PixelFormat32bppRGB:
@@ -892,7 +892,7 @@ GdipCreateBitmapFromScan0 (int width, int height, int stride, PixelFormat format
 #if WORDS_BIGENDIAN
 		int			i;
 #endif
-        
+
 		palette_entries = 1 << gdip_get_pixel_format_depth(format);
 		header_size = sizeof(ColorPalette) - sizeof(ARGB);
 		bytes_needed = header_size + palette_entries * sizeof(ARGB);
@@ -1440,7 +1440,7 @@ gdip_pixel_stream_has_next (StreamingState *state)
 	if (state != NULL) {
 		return (state->p >= 0)
 		    || ((state->y < (state->region.Y + state->region.Height))
-                    && (state->x < (state->region.X + state->region.Width)));
+		    && (state->x < (state->region.X + state->region.Width)));
 	} else {
 		return FALSE;
 	}
--- a/src/font.c
+++ b/src/font.c
@@ -778,7 +778,7 @@ gdip_get_fontfamily_details (GpFontFamily *family, FontStyle style)
 		cairo_scaled_font_t* scaled_ft;
 		FT_Face face = NULL;
 		cairo_matrix_t matrix1, matrix2;
-	        cairo_font_options_t *options = cairo_font_options_create ();
+			cairo_font_options_t *options = cairo_font_options_create ();
 		cairo_font_face_t* cairofnt = gdip_get_cairo_font_face (font);
 
 		cairo_matrix_init (&matrix1, 1, 0, 0, 1, 0, 0);
--- a/src/general.c
+++ b/src/general.c
@@ -104,27 +104,27 @@ GetLibgdiplusVersion ()
 GpStatus 
 gdip_get_status (cairo_status_t status)
 {
-        if (status == CAIRO_STATUS_SUCCESS)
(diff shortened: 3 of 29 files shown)


src/bitmap.c
```diff
@@ -887,7 +887,7 @@ GdipCreateBitmapFromScan0 (int width, int height, int stride, PixelFormat format
 #if WORDS_BIGENDIAN
 		int			i;
 #endif
-        
+
```
akoeplinger: in cases like this there should just be an empty line, not one with tabs.

src/bitmap.c
```diff
@@ -1434,8 +1434,8 @@ gdip_pixel_stream_has_next (StreamingState *state)
 {
 	if (state != NULL) {
 		return (state->p >= 0)
-		    || ((state->y < (state->region.Y + state->region.Height))
-                    && (state->x < (state->region.X + state->region.Width)));
+			|| ((state->y < (state->region.Y + state->region.Height))
+					&& (state->x < (state->region.X + state->region.Width)));
```
akoeplinger: indentation went wrong here.

src/bitmap.c
```diff
@@ -1470,8 +1470,8 @@ gdip_pixel_stream_get_next (StreamingState *state)
 			state->y++;
 
 			state->scan = (BYTE*)(state->data->scan0)
-			            + state->y * state->data->stride
-			            + state->x;
+						+ state->y * state->data->stride
+						+ state->x;
```
akoeplinger: and here

src/bitmap.c
```diff
@@ -1510,8 +1510,8 @@ gdip_pixel_stream_get_next (StreamingState *state)
 			state->y++;
 
 			state->scan = (BYTE*)(state->data->scan0)
-			            + state->y * state->data->stride
-			            + state->x * gdip_get_pixel_format_bpp (state->data->pixel_format) / 8;
+						+ state->y * state->data->stride
+						+ state->x * gdip_get_pixel_format_bpp (state->data->pixel_format) / 8;
```
akoeplinger: and here. I'll stop commenting on these, I think we should just revert cases where the line is indented with tabs and spaces are just used for alignment.

src/font.c
```diff
@@ -775,7 +776,7 @@ gdip_get_fontfamily_details (GpFontFamily *family, FontStyle style)
 		cairo_scaled_font_t* scaled_ft;
 		FT_Face face = NULL;
 		cairo_matrix_t matrix1, matrix2;
-	        cairo_font_options_t *options = cairo_font_options_create ();
+			cairo_font_options_t *options = cairo_font_options_create ();
```
akoeplinger: one tab too much (there are a few cases in this patch).

src/general.c
```diff
@@ -104,27 +104,27 @@ GetLibgdiplusVersion ()
 GpStatus 
 gdip_get_status (cairo_status_t status)
 {
-        if (status == CAIRO_STATUS_SUCCESS)
-                return Ok;
+	if (status == CAIRO_STATUS_SUCCESS)
+			return Ok;
```
akoeplinger: one tab too much.

src/gifcodec.c
```diff
@@ -422,9 +422,9 @@ gdip_load_gif_image (void *stream, GpImage **image, BOOL from_file)
 		si = gif->SavedImages[i];
 		img_desc = &si.ImageDesc;
 		if (img_desc->Top < 0 || img_desc->Height < 0 ||
-		    img_desc->Left < 0 || img_desc->Width < 0 ||
-		    (img_desc->Width + img_desc->Left) > screen_width ||
-		    (img_desc->Height + img_desc->Top) > screen_height) {
+			img_desc->Left < 0 || img_desc->Width < 0 ||
+			(img_desc->Width + img_desc->Left) > screen_width ||
+			(img_desc->Height + img_desc->Top) > screen_height) {
```
akoeplinger: good example of where we should just revert to the previous version with tab + space for alignment. (essentially every change in this file)

src/image.c
```diff
@@ -1040,7 +1040,7 @@ GdipSaveImageToStream (GpImage *image, void *stream, GDIPCONST CLSID *encoderCLS
 	if (!stream || !image)
 		return InvalidParameter;
 
-	return NotImplemented; /* GdipSaveImageToStream - not supported */
```
akoeplinger: restore comment

src/image.c
```diff
@@ -2274,10 +2274,10 @@ GpStatus WINGDIPAPI
 GdipLoadImageFromDelegate_linux (GetHeaderDelegate getHeaderFunc,
 								 GetBytesDelegate getBytesFunc,
 								 PutBytesDelegate putBytesFunc,
-                                 SeekDelegate seekFunc,
+								 SeekDelegate seekFunc,
 								 CloseDelegate closeFunc,
 								 SizeDelegate sizeFunc,
-                                 GpImage **image)
+								 GpImage **image)
```
akoeplinger: it'd probably make more sense to align with GetHeaderDelegate instead.

hughbe:
I've updated the PR to revert the trailing whitespace changes and the comments you made

PR #123, opened 2017-11-14
Fix some memory leaks in codecs if allocations fail

hughbe:


src/emfcodec.c
```diff
@@ -280,6 +283,15 @@ PolyPolygon (MetafilePlayContext *context, BYTE *data, BOOL compact)
 		current->num = GETDW(DWP(n));
 		n++;
 		current->points = (GpPointF*) GdipAlloc (current->num * sizeof (GpPointF));
+		if (!current->points) {
+			for (j = 0; j < i; j++) {
+				GdipFree (list[j]->points);
```
akoeplinger: this fails to compile

hughbe: Fixed..

PR #124, opened 2017-11-15
Make assertion helpers easier to debug

hughbe:
No longer appear in stack traces etc


hughbe:
This brings up a bunch of warnings (for conversion of float to double). These are legit warnings that I'll fix

PR #133, opened 2017-11-22
Fix stack buffer overflow in teststringformat.c

hughbe:
Found with asan

Also clean up some ifdefs


akoeplinger:
Thanks so much for these PRs and sorry for the delay in reviewing 💖

PR #135, opened 2017-11-23
Fixed issues that prevented visibly drawing many WMF files

akoeplinger:
* Added a new drawing instruction / record (METAFILE_RECORD_DIBSTRETCHBLT) which is substantially the same as METAFILE_RECORD_STRETCHDIBITS
 * Corrected parsing of certain values so that they're treated as signed instead of unsigned, including adding a GETS macro to complement the existing GETW but returning a SHORT instead of a WORD
 * Fixed scaling for drawing to a specified size and improved translation at the same time
 * Corrected handling of the DIB BitmapInfoHandler ImageSize field -- it now calculates the value when compression is not used (Compression field = 0 / RGB)

_Note:_ these are the extracted WMF changes from https://github.com/mono/libgdiplus/pull/28 + cleanup, let's see what CI says 😄

### DIFF
--- a/src/metafile.c
+++ b/src/metafile.c
@@ -599,12 +599,19 @@ gdip_metafile_StretchDIBits (MetafilePlayContext *context, int XDest, int YDest,
 	printf ("\n\t\tClrImportant: %d", lpBitsInfo->bmiHeader.biClrImportant);
 #endif
 	ms.ptr = (BYTE*)lpBitsInfo;
-	ms.size = lpBitsInfo->bmiHeader.biSizeImage;
+	if (lpBitsInfo->bmiHeader.biCompression == 0) { // 0 == RGB 
+		// Per the spec, if compression is RGB ImageSize must be ignored (and it should be zero anyway)
+		// and calculated according to the following formula.
+		ms.size = (((lpBitsInfo->bmiHeader.biWidth * lpBitsInfo->bmiHeader.biPlanes * 
+			lpBitsInfo->bmiHeader.biBitCount + 31) & ~31) / 8) * abs(lpBitsInfo->bmiHeader.biHeight);
+	} else {
+		ms.size = lpBitsInfo->bmiHeader.biSizeImage;
+	}
 	ms.pos = 0;
 	status = gdip_read_bmp_image (&ms, &image, Memory);
 	if (status == Ok) {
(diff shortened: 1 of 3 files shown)


akoeplinger:
@hughbe not sure if you played with the WMF code before but if you did I'd appreciate a review from you :)

src/metafile.c
```diff
@@ -599,12 +599,19 @@ gdip_metafile_StretchDIBits (MetafilePlayContext *context, int XDest, int YDest,
 	printf ("\n\t\tClrImportant: %d", lpBitsInfo->bmiHeader.biClrImportant);
 #endif
 	ms.ptr = (BYTE*)lpBitsInfo;
-	ms.size = lpBitsInfo->bmiHeader.biSizeImage;
+	if (lpBitsInfo->bmiHeader.biCompression == 0) { // 0 == RGB 
+		// Per the spec, if compression is RGB ImageSize must be ignored (and it should be zero anyway)
+		// and calculated according to the following formula.
+		ms.size = (((lpBitsInfo->bmiHeader.biWidth * lpBitsInfo->bmiHeader.biPlanes * 
+			lpBitsInfo->bmiHeader.biBitCount + 31) & ~31) / 8) * abs(lpBitsInfo->bmiHeader.biHeight);
```
hughbe: the thing that is being multiplied by `abs(lpBits...)` needs to be floored according to https://en.m.wikipedia.org/wiki/BMP_file_format#Pixel_Storage

src/metafile.c
```diff
@@ -599,12 +599,19 @@ gdip_metafile_StretchDIBits (MetafilePlayContext *context, int XDest, int YDest,
 	printf ("\n\t\tClrImportant: %d", lpBitsInfo->bmiHeader.biClrImportant);
 #endif
 	ms.ptr = (BYTE*)lpBitsInfo;
-	ms.size = lpBitsInfo->bmiHeader.biSizeImage;
+	if (lpBitsInfo->bmiHeader.biCompression == 0) { // 0 == RGB
```
hughbe: Can use BI_RGB macro here

akoeplinger:
@hughbe fixed, thanks!

hughbe:
Looks great thanks! I don’t understand the rest of the #28 changes that aren’t related to metafiles

PR #136, opened 2017-11-23
Add 4bpp BMP codec tests

hughbe:


src/bmpcodec.c
```diff
@@ -155,7 +155,7 @@ gdip_bitmap_fill_info_header (GpBitmap *bitmap, PBITMAPINFOHEADER bmi)
 #endif
 }                                                           
 
-static GpStatus
+static void
 gdip_read_bmp_rle_8bit (void *pointer, BYTE *scan0, BOOL upsidedown, int stride, int scanWidth, int scanCount, ImageSource source)
```
hughbe: GDI+ (and mspaint etc. etc.) silently ignore failures in reading rle8/rle4 bitmaps

Could be worth keeping these checks but just removing the unecessary TODOs

akoeplinger: Could we just ignore the "error" at the call sites and keep the checks here?

hughbe: Yeah - I agree. I reverted this change. I'll fix the TODOs when I fix the crasher bug in the future

hughbe:
Also fixed the test failures...

PR #146, opened 2017-11-30
Fix microsoft code analysis warnings

hughbe:


src/pngcodec.c
```diff
@@ -495,7 +495,6 @@ gdip_load_png_image_from_file_or_stream (FILE *fp, GetBytesDelegate getBytesFunc
 			case 1:
 				if (bit_depth == 2) {
 					/* Make sure transparency is respected for 2bpp images. */
-					memset(alpha, 0xFF, 4);	/* default to fully opaque */
```
akoeplinger: please add that comment to the variable definition above

src/jpegcodec.c
```diff
@@ -477,12 +477,14 @@ gdip_load_jpeg_image_internal (struct jpeg_source_mgr *src, GpImage **image)
 					 * looks like BGR data. */
 					inptr -= 3;
 					outptr -= 4;
+					g_assert (outptr);
```
akoeplinger: should we assert inptr too?

src/jpegcodec.c
```diff
@@ -477,12 +477,14 @@ gdip_load_jpeg_image_internal (struct jpeg_source_mgr *src, GpImage **image)
...
 					set_pixel_bgra(outptr, 0, inptr[2], inptr[1], inptr[0], 0xff);
 				}
 				/* keep last 2 lines in temporary variables */
 				if (width > 1) {
 					BYTE b2, g2, r2, b1, g1, r1;
 					inptr -= 3;
+					g_assert (inptr);
```
akoeplinger: inptr is asserted here but not a few lines below, why?

PR #148, opened 2017-12-11
Make image verification failures easier to spot

hughbe:
Before there would be no line number for the actual data that caused the verification to fail

We can fix this by turning the function into a macro

Relies on #147


akoeplinger:
This needs a rebase now that I merged the other PR :)

hughbe:
Rebased, thanks

PR #151, opened 2017-12-13
Handle infinite recursive directories in tiff files

hughbe:


directhex:
I had to abort the builds on Jenkins. It was hanging forever. Maybe some kind of infinite recursion issue.........

hughbe:
This passes on mac/windows but not CI Linux. Maybe libtiff is outdated?

I know that they fixed the recursive function in libtiff:
https://github.com/vadz/libtiff/commit/0417e415288bd4af92a534e040be23d444b2d2d5
https://github.com/vadz/libtiff/commit/bcc3de6d25ddcb7e62191515c2155c5d46d37d2c

akoeplinger:
Yeah we won't have a libtiff from 2015 on all platforms so we'll need to find a workaround.

hughbe:
Eeek @directhex I must've got the fix wrong... again :// you might need to terminate the build

akoeplinger:
Aborted and added a 30mins timeout to the jobs :)

hughbe:
I tried to create our own version of `TiffNumberOfDirectories` to prevent this infinite recursion, but there was no good way to make it work as `TiffNumberOfDirectories` uses internal unexposed helpers as well as internal data structures I can't access.

The best thing I could do was to disable the test for versions of libtiff older that 4.0.4 to prevent infinite recursion there. 

This is the best fix I can think of, but still can cause denial of service on older libtiff machines. The infinite recursion is fixed with newer machines though, as we now error out if the number of directories is `65535`

PR #186, opened 2017-12-17
Support later versions of BITMAPINFOHEADER and fix 16bpp decoding

### LINKED ISSUE #187: Disallow unknown biSize values of BITMAPINFOHEADER
Seems like GDI+ rejects these, currently we don't. Need to add tests

### LINKED ISSUE #188: Support BITMAPV3/V4/V5HEADER for bitmaps
We currently just assume that if the header is not OS/2, then it must be BITMAPINFOHEADER which is wrong. Need to add tests for these headers also verifying the pixel values of the image, as previously we'd read data from the palette or header as image data

hughbe:
And disallow unknown sizes of headers to match GDI+ behaviour

Previously, we would read only 40 bytes worth of the header at all times, whatever the biSize field of the header said.

This means that we'd interpret bytes from BITMAPV3/V4/V5HEADER as palette or image data

For 16bpp we would cause a heap buffer overflow/crash and this fixes that bub too

Fixes #187 
Fixes #188


hughbe:
@akoeplinger sorry to be annoying but here's a friendly ping to check if you're able to review my open PRs? I'd like to contribute a bit more soon so getting rid of the backlog would be great

src/bmpcodec.c
```diff
@@ -745,39 +770,84 @@ gdip_read_BITMAPINFOHEADER (void *pointer, BITMAPINFOHEADER *bmi, ImageSource so
...
+		return OutOfMemory;
+	((BITMAPV3HEADER *) bmi)->bV3AlphaMask = (data_read[3]<<24 | data_read[2]<<16 | data_read[1]<<8 | data_read[0]);
+
+	/* We don't use any information in BITMAPV4HEADER or BITMAPV5HEADER, so there is no point going through all of
+	 * the other fields. This leaves the rest of the structure uninitialized. */
+	size = headerSize - sizeof (BITMAPV3HEADER);
+	if (size > 0) {
+		while (size > sizeof (DWORD)) {
+			if (gdip_read_bmp_data (pointer, data_read, sizeof (DWORD), source) != sizeof (DWORD))
+				return OutOfMemory;
```
filipnavara: Shouldn't `bmi` be deallocated here?

src/bmpcodec.c
```diff
@@ -745,39 +770,84 @@ gdip_read_BITMAPINFOHEADER (void *pointer, BITMAPINFOHEADER *bmi, ImageSource so
...
+			size -= sizeof (DWORD);
+		}
+		if (size > 0) {
+			if (gdip_read_bmp_data (pointer, data_read, size, source) != size)
+				return OutOfMemory;
```
filipnavara: Shouldn't `bmi` be deallocated here too?

src/bmpcodec.c
```diff
@@ -1068,13 +1140,19 @@ gdip_read_bmp_image (void *pointer, GpImage **image, ImageSource source)
...
+	if (bmi->biXPelsPerMeter != 0 && bmi->biYPelsPerMeter != 0)
 		result->active_bitmap->image_flags |= ImageFlagsHasRealDPI;
 
+	GdipFree (bmi);
+
 	*image = result;
 	return Ok;
 
 error:
+	if (bmi != NULL) {
```
hughbe: @filipnavara bmi is always deallocated here and above

filipnavara: Fair enough, I missed that. I am not a big fan of functions that return failure and require additional cleanup. It would seem cleaner to me if the cleanup was done directly in gdip_read_BITMAPINFOHEADER.

akoeplinger:
@hughbe my apologies, I screwed up my email filter rules for GitHub so I didn't get any pings from this repository...

I'll start working my way through the PR list :)

PR #196, opened 2017-12-19
Support giflib on Windows

### LINKED ISSUE #194: Support giflib on windows
This is the only missing library on Windows libgdiplus

I can't see a nuget package anywhere, maybe I'll create one :D

hughbe:
I don't really understand why we need the FreeRasterBitsMono workaround. The problem is that we crash calling DGifCloseFile when freeing `RasterBits`.

For some reason, freeing it manually works. This might need more investigation, but I'm at a loss here

Fixes #194

### DIFF
--- a/src/libgdiplus.vcxproj
+++ b/src/libgdiplus.vcxproj
@@ -91,7 +91,6 @@
       <PreprocessorDefinitions>HAVE_LIBGIF;HAVE_LIBJPEG;HAVE_LIBTIFF;HAVE_LIBPNG;HAVE_FCFINI;_WINDLL;WIN32;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <!-- FIXME: To align with the GDI+ calling convention, this should be StdCall. Only relevant on x86 -->
       <CallingConvention>Cdecl</CallingConvention>
-      <RuntimeLibrary>MultiThreadedDLL</RuntimeLibrary>
     </ClCompile>
     <Link>
       <AdditionalDependencies>libpng16.lib;freetype.lib;fontconfig.lib;glib-2.0.lib;cairo.lib;shcore.lib;%(AdditionalDependencies)</AdditionalDependencies>
--- a/src/libgdiplus.vcxproj
+++ b/src/libgdiplus.vcxproj
@@ -88,7 +88,7 @@
   <ItemDefinitionGroup>
     <ClCompile>
       <AdditionalIncludeDirectories>$(ProjectDir)\..;$(ProjectDir)\..\gtk\$(Platform)\release\include;$(ProjectDir)\..\gtk\$(Platform)\release\include\glib-2.0;$(ProjectDir)\..\gtk\$(Platform)\release\lib\glib-2.0\include;$(ProjectDir)\..\gtk\$(Platform)\release\include\cairo;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
-      <PreprocessorDefinitions>HAVE_LIBJPEG;HAVE_LIBTIFF;HAVE_LIBPNG;HAVE_FCFINI;_WINDLL;WIN32;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions>HAVE_LIBGIF;HAVE_LIBJPEG;HAVE_LIBTIFF;HAVE_LIBPNG;HAVE_FCFINI;_WINDLL;WIN32;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <!-- FIXME: To align with the GDI+ calling convention, this should be StdCall. Only relevant on x86 -->
       <CallingConvention>Cdecl</CallingConvention>
       <RuntimeLibrary>MultiThreadedDLL</RuntimeLibrary>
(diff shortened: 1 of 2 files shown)


hughbe:
So I did a bit of digging and found out that the following code crashes on Windows:

```cpp 
void *data = GdipAlloc(1);
free(data);
```

But replacing free with GdipFree works. I think the problem is that the call to Malloc is in Libgdiplus code but the call to free is in libgif

@qmfredrik FYI as you did most of the work getting libgdiplus compiling on Windows. Do you have any ideas?

qmfrederik:
@hughbe I'm no C++ master but I do believe you must use the same library to allocate and free memory. That is, if you use `malloc` you should use `free`, and if you use `GdipAlloc` you should use `GdipFree`.

Some libraries (such as libpng) allow you to specify which functions they use to allocate/free memory, it doesn't look like giflib allows you to do that.

Net, if you pass memory to giflib which it will "own" (and free), I think the best way is to make sure that memory has been allocated with malloc. If that isn't possible, this workaround should do the trick, too.

hughbe:
@qmfrederik hi, do you have time to review this change? I’d like to get some of my open prs closed so I can make some more changes without having to get confused with rebases  etc

qmfrederik:
@hughbe Did you mean to ping @akoeplinger? This change looks good to me but I'm not on the mono/libgdiplus team.

That aside, I'd love for your patches to land in libgdplus as well 😄 .

hughbe:
derp!! my bad

filipnavara:
It's perfectly true that you cannot mix malloc/GdipAlloc with GdipFree/free with gdiplus.dll. Inside libgdiplus it doesn't really matter though because GdipAlloc is implemented using malloc. There's nothing wrong with making the change though.

src/gifcodec.c
```diff
@@ -207,7 +207,7 @@ DGifSlurpMono(GifFileType * GifFile, SavedImage *TrailingExtensions)
 				if (ImageSize == 0)
 					return GIF_ERROR;
 
-				sp->RasterBits = (BYTE*) GdipAlloc (ImageSize * sizeof (GifPixelType));
+				sp->RasterBits = (BYTE*) malloc (ImageSize * sizeof (GifPixelType));
```
akoeplinger: This doesn't make sense to me, GdpiAlloc is just a wrapper around malloc:

https://github.com/mono/libgdiplus/blob/63aed45cc6d4c10c02cb02e3cd1256d0317ddf93/src/general.c#L72-L76

akoeplinger:
@filipnavara the problem I have with this patch is that it just replaces one of the malloc() in the file while there are dozen other GdipAlloc() left. We should first understand why it's crashing since there shouldn't be a difference between malloc and GdipAlloc inside of our own libgdiplus.dll

filipnavara:
Sure, I agree. I cannot guess what was @hughbe original intention.

What I wanted to say is that the field is part of giflib API contract and it specified that it is allocated using malloc:

```
typedef struct SavedImage {
    GifImageDesc ImageDesc;
    GifByteType *RasterBits;         /* on malloc(3) heap */
    int ExtensionBlockCount;         /* Count of extensions before image */    
    ExtensionBlock *ExtensionBlocks; /* Extensions before image */    
} SavedImage;
```

All the other allocations are made in such a manner that the caller is also responsible for the deallocation and thus any heap can be used.

hughbe:
Okay I've looked into this further:

We crash with the following message in Debug (and hang in release)
error

We crash decoding the gif:

```cpp
     ...
	FreeExtensionMono(&global_extensions);
#if (GIFLIB_MAJOR > 5) || ((GIFLIB_MAJOR == 5) && (GIFLIB_MINOR >= 1))
	DGifCloseFile (gif, NULL); // HERE
#else
	DGifCloseFile (gif);
#endif
     ...
```

We crash in `DGifCloseFile`:
```cpp
    ...
    if (GifFile->SavedImages) {
        GifFreeSavedImages(GifFile); // HERE
        GifFile->SavedImages = NULL;
    }
    ...
```

Inside `GifFreeSavedImages`:
```cpp
void
GifFreeSavedImagesMono(GifFileType *GifFile)
{
    SavedImage *sp;

    if ((GifFile == NULL) || (GifFile->SavedImages == NULL)) {
        return;
    }
    for (sp = GifFile->SavedImages;
         sp < GifFile->SavedImages + GifFile->ImageCount; sp++) {
        if (sp->ImageDesc.ColorMap != NULL) {
            GifFreeMapObject(sp->ImageDesc.ColorMap);
            sp->ImageDesc.ColorMap = NULL;
        }

        if (sp->RasterBits != NULL)
            free((char *)sp->RasterBits);
	
	GifFreeExtensions(&sp->ExtensionBlockCount, &sp->ExtensionBlocks);
    }
    free((char *)GifFile->SavedImages); // HERE
    GifFile->SavedImages = NULL;
}
```

I don't actually know how to fix this one. It just doesn't make sense!

src/libgdiplus.vcxproj
```diff
@@ -88,7 +88,7 @@
   <ItemDefinitionGroup>
     <ClCompile>
       <AdditionalIncludeDirectories>$(ProjectDir)\..;$(ProjectDir)\..\gtk\$(Platform)\release\include;$(ProjectDir)\..\gtk\$(Platform)\release\include\glib-2.0;$(ProjectDir)\..\gtk\$(Platform)\release\lib\glib-2.0\include;$(ProjectDir)\..\gtk\$(Platform)\release\include\cairo;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
-      <PreprocessorDefinitions>HAVE_LIBJPEG;HAVE_LIBTIFF;HAVE_LIBPNG;HAVE_FCFINI;_WINDLL;WIN32;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions>HAVE_LIBGIF;HAVE_LIBJPEG;HAVE_LIBTIFF;HAVE_LIBPNG;_WINDLL;WIN32;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
filipnavara: Merge error here :-)

kasper3:
@hughbe, I have built your branch using VS2017 on Windows 10. Can you please share the TestGdiplus project so I can debug the invalid heap ptr issue?

Aside, I had to make lots of small tweaks in packages directory to make it compile on VS2017, where I don't have VS2015 installed. Which makes me wonder:

1. Why are there toolset directories v120, v1.40 for binaries under packages? If we compile with latest v141 toolset and have WindowsXP flag on, the resultant dll is compatible with Windows XP -> Windows 10, so why bother asserting on toolset?
2. vcxproj project name should be agnostic of VS2015? When you open SLN in VS2017-only environment on Windows 10. It should just build. We can set detection of Windows 10 SDK version available using inline function in vcxproj:

  ```xml
      <WindowsTargetPlatformVersion Condition="'$(WindowsTargetPlatformVersion)' == ''">$([Microsoft.Build.Utilities.ToolLocationHelper]::GetLatestSDKTargetPlatformVersion('Windows', '10.0'))</WindowsTargetPlatformVersion>
  ```

hughbe:
@kasper3 download this file and put it in the directory that contains the `libgdiplus` folder that you cloned from github

[TestGdiplus.zip](https://github.com/mono/libgdiplus/files/1865790/TestGdiplus.zip)

kasper3:
@hughbe, so I was able to reproduce the assertion. Then I took the giflib sources from https://github.com/rcancro/giflib/releases/tag/5.1.2 and added to the project (adjusted some headers like unistd.h in giflib sources), and built, error went away. It seems like we have issue with the dll that is packed in the nuget. Maybe try releasing the nuget with latest VS2017 and latest compiler toolchain (via AppVeyor CI for example)?

hughbe:
I think I have it!!! Nothing is wrong with the dll. It's that we're always using the `MultiThreadedDLL` debug library, rather than the `MultiThreadedDebugDLL` runtime library in libgdiplus. Giflib, however, uses the correct runtime library on Debug mode.

However, there's a mismatch between the two which causes this error!

kasper3:
This looks pretty good. 👍 

@hughbe, btw, where did you get a working `fonts.conf`? I didn't have it at `%WINDIR%/fonts.conf`, so I tried using one in `gtk\Win32\release\etc\fonts`, that resulted in error like `cannot have <blank> in conf`, so I removed that block out from xml.

PR #198, opened 2017-12-19
Fix crash in GdipPrivateAddMemoryFont on Windows

hughbe:
write(3) doesn't accept a FILE* pointer masquerading as an int


filipnavara:
On a second thought, couldn't we keep the code as is and just modify CreateTempFile to call "open" at the end instead of "fopen"?

akoeplinger:
@filipnavara maybe, but I don't really care in this case :)

filipnavara:
Neither do I, honestly, as long as it works.

PR #201, opened 2017-12-20
Implement gdip_get_display_dpi on Windows

hughbe:


filipnavara:
This unnecessarily restricts the code to work only on Windows 8.1+. Since the per-monitor information is not used I would suggest to use the good old [GetDeviceCaps](https://msdn.microsoft.com/en-us/library/windows/desktop/dd144877(v=vs.85).aspx) API instead. The LOGPIXELSX and LOGPIXELSY values should be what you need.

akoeplinger:
@filipnavara we're already targeting 8.1 in the .vcxproj so this should be fine for now.

filipnavara:
Fair enough, I will submit a PR myself. It's a bit of a pet peeve for me since I regularly have to work on computers with older SDK headers and open source compilers like MinGW.

PR #205, opened 2017-12-21
Copy missing dlls on Windows

### LINKED ISSUE #203: Referencing x64 Windows causes crash on load for bad image
It looks like we’re linking with a 32bit dll (pixman-1.dll)

This causes a crash like in https://stackoverflow.com/questions/28299246/0xc000007b-error

hughbe:
Fixes #203


akoeplinger:
These libraries don't seem to be available in the new gtk+ package from https://github.com/mono/libgdiplus/commit/38e739733237688677d15dc07fca5671e54da39b

PR #220, opened 2018-01-21
Fix no error when the font collection is null and the font doesn't exist

### LINKED ISSUE #161: GdipCreateFontFamilyFromName does not validate that the font exists if there is no collection
In `test_createFontFamilyFromName` in `testfont.c`

```
	// FIXME: Libgdiplus does not validate that the font family exists
	// if the collection is NULL.
#if defined(USE_WINDOWS_LIBGDIPLUS)
	status = GdipCreateFontFamilyFromName (NoSuchFont, NULL, &family);
	assertEqualInt (status, FontFamilyNotFound);
#endif
```

hughbe:
And delete dead code as a result

Fixes #161

### DIFF
--- a/src/font-private.h
+++ b/src/font-private.h
@@ -57,8 +57,6 @@ struct _Font {
 #endif
 };
 
-void gdip_font_clear_pattern_cache (void) GDIP_INTERNAL;
-
 #ifdef USE_PANGO_RENDERING
 
 PangoFontDescription* gdip_get_pango_font_description (GpFont *font);
--- a/src/font.c
+++ b/src/font.c
@@ -327,132 +327,6 @@ gdip_status_from_fontconfig (FcResult result)
 	}
 }
 
-/* note: MUST be executed inside a lock because FcConfig isn't thread-safe */
-static FcPattern*
-create_pattern_from_name (char* name)
-{
-	FcValue val;
-	/* FcResult must be initialized because it's changed only in error conditions */
-	FcResult rlt = FcResultMatch;
-	FcPattern *full_pattern = NULL;
-	FcPattern *name_pattern = FcPatternCreate ();
-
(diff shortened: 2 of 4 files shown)


hughbe:
Relies on https://github.com/mono/libgdiplus/pull/221

filipnavara:
The added test fails on OS X and Linux.

I am a bit torn on this one. On one hand I would very much like the mutexes gone and the font matching code unified. On the other hand I am worried that this will result in a performance regression.

filipnavara:
The test actually fails because now the GdipGetGenericFontFamilySansSerif API fails. There's no "MS Sans Serif" on macOS by default.

filipnavara:
I have added font substitution with FcConfigSubstitute and FcDefaultSubstitute to create_fontfamily_from_collection. Now it maps "MS Sans Serif" to "Bitstream Vera Sans" and still fails to find it. :-)

hughbe:
I can't seem to figure this one out, so I'm gonna close it. Feel free to make an attempt fixing this differently if you want

filipnavara:
It boils down to the fact that the available fonts on different platform are different. Since the main purpose of libgdiplus is to offer compatibility with existing code which is likely to use standard Windows font names there has to be some sort of mapping done. There are basically three ways how to do it:

- Use whatever FontConfig does and never fail to find a font
- Use your approach along with fallback mapping for common fonts on each supported platform (Windows, macOS, Linux w/ FreeDesktop base, any others?)
- Ship the Microsoft or Wine fonts as a fallback mechanism

The second approach is hard to maintain and requires us to do per-platform changes. The benefits may be negligible unless there's a compelling use case for the behavior. The third approach could be a nightmare when it comes to packaging and licensing requirements.

Issue #224, opened 2018-03-15
GdipCreatePath leaks memory in error case


filipnavara:
If the second allocation in GdipCreatePath fails it leaks memory from the previous allocations.

filipnavara:
Same bug in GdipClonePath.

hughbe:
Should be closed, fixed by #227

PR #226, opened 2018-03-16
Fix memory leak in error case of GdipCreatePath

hughbe:
Fixes https://github.com/mono/libgdiplus/issues/224


filipnavara:
Thanks! I started working on a more ambitious PR, but it fails test at the moment, so it may be easier to just fix the small things first.

hughbe:
Closing in favour of your PR

filipnavara:
I will borrow you your test if you don't mind.

Issue #223, opened 2018-03-15
GdipGetPathData memory managment


filipnavara:
It seems that the implementation of GdipGetPathData is pretty wrong in relation to how it manages memory. The caller is supposed to allocate the buffers and specify the point count. The current implementation leaks memory on each call by allocating buffers and overriding the passed buffers.

Reference: https://referencesource.microsoft.com/#System.Drawing/commonui/System/Drawing/Advanced/GraphicsPath.cs,321

hughbe:
Should be closed, fixed by #227a

Issue #225, opened 2018-03-15
GdipGetPathPointsI doesn't check for count > path->count


filipnavara:
This missing error check could cause buffer overrun.

hughbe:
Should be closed, fixed by #227

PR #227, opened 2018-03-16
Rewritten memory management of GpPath objects

filipnavara:
* Fixes issues #223, #224 and #225.
* Introduces out of memory checks for the path APIs.
* Optimises lot of the APIs to prevent unnecessary allocations, frees and copying.

### DIFF
--- a/src/customlinecap.c
+++ b/src/customlinecap.c
@@ -212,8 +212,8 @@ gdip_custom_linecap_draw (GpGraphics *graphics, GpPen *pen, GpCustomLineCap *cus
 
 		for (i = 0; i < points; i++) {
 			/* Adapted from gdip_plot_path() */
-			GpPointF point = g_array_index (path->points, GpPointF, i);
-			BYTE type = g_array_index (path->types, BYTE, i);
+			GpPointF point = path->points[i];
+			BYTE type = path->types[i];
 			GpPointF pts [3];
 
 			/* mask the bits so that we get only the type value not the other flags */
--- a/src/graphics-cairo.c
+++ b/src/graphics-cairo.c
@@ -425,8 +425,8 @@ gdip_plot_path (GpGraphics *graphics, GpPath *path, BOOL antialiasing)
 	int i, idx = 0;
 
 	for (i = 0; i < length; ++i) {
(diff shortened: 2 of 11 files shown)


akoeplinger:
@filipnavara is there a particular reason for "Remove G[Byte]Array code from GpPath objects and optimize the code"? It is touching a lot of code that I'm not familiar with :)

filipnavara:
There's actually a couple of reasons, some of them objective, other subjective:

1. The G[Byte]Array have no way of handling out-of-memory situation gracefully. They just enter an infinite loop and freeze the application.
2. The g_array_append_val makes it easy to do a lot of reallocations resulting in an suboptimal code.
3. Old code actually did a LOT of unnecessary allocations/reallocations/copying.
4. I was fed up with glib dependency on Win32 / OS X and have a glib-free version of the code.

Admittedly the only objective reason is number 1. Number 2 and 3 could be solved while keeping the g_array_*, but it would result in a similarly big patch. Number 4 is moot for now, but maybe there would be some interest in that eventually.

filipnavara:
I tested the code with libgdiplus test, Mono tests and KeePass running on OS X.

akoeplinger:
Ok makes sense. Could you please rebase/squash the commits so we only have these:

- Remove G[Byte]Array code from GpPath objects and optimize the code
- Check for out-of-memory errors. Add missing copying in GdipCreatePath2.
- Fix memory leak in GdipFlattenPath and out-of-memory handling.
- Initial unit test for path functions.

kasper3:
This is a great progress!
@filipnavara, is it towards dropping the dependency on glib?

filipnavara:
@kasper3 I am no longer sure whether dropping GLib dependency is worth it. Yes, this patch goes a long way towards it. The remaining stuff is easy to rewrite too. Then again, if we decide to make use of Pango text rendering there will be an indirect dependency anyway.

filipnavara:
I will do another rebase after https://github.com/mono/mono/pull/7943 is merged. Otherwise this is ready to be merged.

hughbe:
What’s the status on this PR? Unfortunately if a PR adds a test file it will conflict with all others that add new test files so would be nice to merge these path test

filipnavara:
Rebased.

PR #228, opened 2018-03-19
Implement gdip_get_display_dpi on Windows using older Win32 API.

filipnavara:


akoeplinger:
@monojenkins whitelist

src/general.c
```diff
@@ -202,17 +197,10 @@ gdip_get_display_dpi ()
...
-		/* Read the dpi from the main monitor, or use a a default of 96.0f. */
-		monitor = MonitorFromWindow (NULL, MONITOR_DEFAULTTOPRIMARY);
-		res = GetDpiForMonitor (monitor, MDT_DEFAULT, &dpiX, &dpiY);
-		if (res == S_OK)
-			dpis = dpiX;
-		else
-			dpis = 96.0f;
+		HDC dc;
+		dc = GetDC (0);
+		dpis = GetDeviceCaps (dc, LOGPIXELSX);
```
akoeplinger: LOGPIXELSX seems to not be available during the CI build, guess we're missing some include:

```
00:21:29   d:\j\workspace\test-libgdiplus-pull-request-windows\src\general.c(202): error C2065: 'LOGPIXELSX': undeclared identifier [d:\j\workspace\test-libgdiplus-pull-request-windows\src\libgdiplus.vcxproj]
00:21:29 
00:21:29     0 Warning(s)
00:21:29     1 Error(s)
```

filipnavara: I guess it's the "#define NOGDI" in front of "#include <Windows.h>" in win32structs.h. I can just place the additional "#define LOGPIXELSX 88" there, but it makes me wonder why it doesn't complain about the function calls.

filipnavara: Just checked that the function prototypes are wrapped in the same NOGDI #ifndef in wingdi.h. This will require some bigger refactoring to get it working. I will look into removing the NOGDI and restructuring win32structs.h.

akoeplinger: I'd be fine with a `GetDeviceCaps (dc, 88 /* LOGPIXELSX */);` for now

filipnavara: I'll triple check the function prototypes first because I am worried that it may be falling back to C behavior with default prototypes and missing the STDCALL calling convention. The relevant warning doesn't seem to be enabled for the Windows builds.

filipnavara:
I am not particularly happy about this PR, so comments are welcome. Right now it removes the "#define NOGDI" for Windows headers and resolves all the conflicts caused by that. The unfortunate side effect is that it is more prone to be broken by name clashes when someone develops on non-Windows systems and chooses to name a function with same name as existing Win32 API. It will be caught by the CI, but could be annoying nonetheless.

filipnavara:
...and I should definitely find a way to enable more warning in the Windows builds to spot calls to undeclared functions.

filipnavara:
Another alternative approach is to go the Carbon route, create a separate .c file with `gdip_get_display_dpi_win32` and #include <windows.h> there.

PR #231, opened 2018-03-20
[DO NOT MERGE] Fix no error when the font collection is null and the font doesn't exist

filipnavara:
This is experimental rework of https://github.com/mono/libgdiplus/pull/220. It doesn't work on OS X, but I want to check if the font substitution makes any difference on Linux.

### DIFF
--- a/src/font-private.h
+++ b/src/font-private.h
@@ -57,8 +57,6 @@ struct _Font {
 #endif
 };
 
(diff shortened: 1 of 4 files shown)


filipnavara:
Seems like the font substitution doesn't solve anything even on Linux. 👎

PR #233, opened 2018-03-22
Transform path in GdipSetClipPath by clip_matrix if necessary.

filipnavara:
- Possible fix for https://github.com/dotnet/corefx/issues/27320.
- Needs unit tests.
- The metafile code doesn't make sense, I am aware of that. It is copied from `GdipSetClipRegion` and the `metafile_SetClip*` functions are not implemented anyway.

### DIFF
--- a/src/graphics.c
+++ b/src/graphics.c
@@ -2024,23 +2024,47 @@ GpStatus WINGDIPAPI
 GdipSetClipPath (GpGraphics *graphics, GpPath *path, CombineMode combineMode)
 {
 	GpStatus status;
+	GpPath *work;
 
 	if (!graphics || !path)
 		return InvalidParameter;
 
-	status = GdipCombineRegionPath (graphics->clip, path, combineMode);	
+	/* if the matrix is empty, avoid path cloning and transform */
+	if (gdip_is_matrix_empty (graphics->clip_matrix)) {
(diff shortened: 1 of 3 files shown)


filipnavara:
In the process of adding the unit test I discovered a bug in `gdip_region_bitmap_get_smallest_rect`. I didn't solve it in the most efficient way, but at least it makes the code behave correctly.

src/graphics.c
```diff
@@ -2024,23 +2024,47 @@ GpStatus WINGDIPAPI
...
-	status = GdipCombineRegionPath (graphics->clip, path, combineMode);	
+	/* if the matrix is empty, avoid path cloning and transform */
+	if (gdip_is_matrix_empty (graphics->clip_matrix)) {
+		work = path;
+	} else {
+		cairo_matrix_t inverted;
+
+		gdip_cairo_matrix_copy (&inverted, graphics->clip_matrix);
+		cairo_matrix_invert (&inverted);
+
```
filipnavara: I actually discovered that there's always an inverted version of `graphics->clip_matrix` in `graphics->copy_of_ctm`. I'd prefer to simplify the code in a separate patch though, along with existing code in `GdipSetClipRect` and `GdipSetClipRegion`.

PR #234, opened 2018-03-27
Fix GDI+ initialization in testfont.

filipnavara:


filipnavara:
Good point, I submitted PR for the other GdiplusStartup calls. The problem was that they were called with invalid parameters, so GdiplusStartup just returned error and no initialization was performed.

PR #238, opened 2018-03-27
Fix initialization of pango font descriptors

filipnavara:


src/font.c
```diff
@@ -705,7 +705,8 @@ PangoFontDescription*
 gdip_get_pango_font_description (GpFont *font)
 {
 	if (!font->pango) {
-		font->pango = pango_font_description_from_string ((char*)font->face);
+		font->pango = pango_font_description_new ();
```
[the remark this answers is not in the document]
filipnavara: This is glib-based code. It never returns NULL. In out-of-memory situation it set the hell on fire and lets your application burn with it.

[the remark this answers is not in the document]
filipnavara: https://github.com/GNOME/pango/blob/b97089825723294cb6d4cc75c6e100c195b84369/pango/fonts.c#L98

src/font.c
```diff
@@ -705,7 +705,8 @@ PangoFontDescription*
...
+		pango_font_description_set_family (font->pango, (char *)font->face);
```
akoeplinger: are you sure this does everything that `pango_font_description_from_string ()` did before? Not sure I fully grok the reason for this change.

filipnavara: It does not, which is kind of the point. `pango_font_description_from_string` does some advanced parsing of the string to support multiple families or size specification. `font->face` is copy of the name in GpFontFamily, which is what the caller really asked for.

akoeplinger: ok :)

PR #243, opened 2018-03-27
Update DGifSlurpMono to match upstream giflib

hughbe:
Rules out our code harming things in https://github.com/mono/libgdiplus/pull/196

This also gives us some additional checks for invalid data!

### DIFF
--- a/src/gifcodec.c
+++ b/src/gifcodec.c
@@ -203,7 +203,7 @@ DGifSlurpMono(GifFileType * GifFile, SavedImage *TrailingExtensions)
 				}
 
 				ImageSize = sp->ImageDesc.Width * sp->ImageDesc.Height;
-				if (ImageSize > (SIZE_MAX / sizeof(GifPixelType))) {
+				if (ImageSize == 0 || ImageSize > (SIZE_MAX / sizeof(GifPixelType))) {
 					return GIF_ERROR;
 				}
 
--- a/src/gifcodec.c
+++ b/src/gifcodec.c
@@ -197,8 +197,7 @@ DGifSlurpMono(GifFileType * GifFile, SavedImage *TrailingExtensions)
 
 				sp = &GifFile->SavedImages[GifFile->ImageCount - 1];
 				/* Allocate memory for the image */
-				if (sp->ImageDesc.Width < 0 && sp->ImageDesc.Height < 0 &&
-						sp->ImageDesc.Width > (INT_MAX / sp->ImageDesc.Height)) {
+				if (sp->ImageDesc.Width < 0 && sp->ImageDesc.Height < 0 && sp->ImageDesc.Width > (INT_MAX / sp->ImageDesc.Height)) {
 					return GIF_ERROR;
 				}
 
@@ -207,7 +206,7 @@ DGifSlurpMono(GifFileType * GifFile, SavedImage *TrailingExtensions)
 					return GIF_ERROR;
 				}
 
-				sp->RasterBits = (unsigned char *) malloc(ImageSize * sizeof(GifPixelType));
+				sp->RasterBits = (BYTE *) GdipAlloc (ImageSize * sizeof(GifPixelType));
 				if (sp->RasterBits == NULL) {
 					return GIF_ERROR;
 				}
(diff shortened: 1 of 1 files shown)


src/gifcodec.c
```diff
@@ -198,40 +190,46 @@ DGifSlurpMono(GifFileType * GifFile, SavedImage *TrailingExtensions)
...
+					* offsets and jumps...
+					*/
+					int InterlacedOffset[] = { 0, 4, 2, 1 };
+					int InterlacedJumps[] = { 8, 8, 4, 2 };
+					/* Need to perform 4 passes on the image */
+					for (i = 0; i < 4; i++) {
+						for (j = InterlacedOffset[i]; j < sp->ImageDesc.Height; j += InterlacedJumps[i]) {
+							if (DGifGetLine(GifFile, sp->RasterBits + j * sp->ImageDesc.Width, sp->ImageDesc.Width) == GIF_ERROR)
+								return GIF_ERROR;
+						}
```
filipnavara: This is essentially the same code as before except for the removed `* sizeof (GifPixelType)`. I found the previous code slightly more readable, but the change seems to be correct.

hughbe: I don't think we should be writing our own code here though - I'd prefer to match upstream giflib rather than roll our own. Although it probably doesn't matter here - I have no strong opinion

filipnavara: No strong opinion either. I didn't realize that it was copied upstream code in the first place. Just wanted to point out what actually changed and could have caused the memory corruption in the first place.

src/gifcodec.c
```diff
@@ -198,40 +191,46 @@ DGifSlurpMono(GifFileType * GifFile, SavedImage *TrailingExtensions)
...
 				sp = &GifFile->SavedImages[GifFile->ImageCount - 1];
+				/* Allocate memory for the image */
+				if (sp->ImageDesc.Width < 0 && sp->ImageDesc.Height < 0 &&
+						sp->ImageDesc.Width > (INT_MAX / sp->ImageDesc.Height)) {
+					return GIF_ERROR;
+				}
+
 				ImageSize = sp->ImageDesc.Width * sp->ImageDesc.Height;
-				if (ImageSize == 0)
+				if (ImageSize > (SIZE_MAX / sizeof(GifPixelType))) {
```
filipnavara: ...and this fails the test on CI (testgifcodec, `createFile (zeroImageWidth, OutOfMemory);`). Maybe `if (ImageSize == 0 || ImageSize > (SIZE_MAX / sizeof(GifPixelType)))` ?

src/gifcodec.c
```diff
@@ -198,40 +191,46 @@ DGifSlurpMono(GifFileType * GifFile, SavedImage *TrailingExtensions)
...
 			case IMAGE_DESC_RECORD_TYPE: {
-				/* This call will leak GifFile->Image.ColorMap; there's a fixme in the DGifGetImageDesc code */
 				if (DGifGetImageDesc(GifFile) == GIF_ERROR) {
 					return (GIF_ERROR);
 				}
 
 				sp = &GifFile->SavedImages[GifFile->ImageCount - 1];
+				/* Allocate memory for the image */
+				if (sp->ImageDesc.Width < 0 && sp->ImageDesc.Height < 0 &&
+						sp->ImageDesc.Width > (INT_MAX / sp->ImageDesc.Height)) {
```
akoeplinger: formatting is weird here, probably better to keep it on a single line.

src/gifcodec.c
```diff
@@ -198,40 +191,46 @@ DGifSlurpMono(GifFileType * GifFile, SavedImage *TrailingExtensions)
...
+				}
+
 				ImageSize = sp->ImageDesc.Width * sp->ImageDesc.Height;
-				if (ImageSize == 0)
+				if (ImageSize == 0 || ImageSize > (SIZE_MAX / sizeof(GifPixelType))) {
 					return GIF_ERROR;
+				}
 
-				sp->RasterBits = (BYTE*) GdipAlloc (ImageSize * sizeof (GifPixelType));
+				sp->RasterBits = (unsigned char *) malloc(ImageSize * sizeof(GifPixelType));
```
akoeplinger: we should still use `GdipAlloc ()`, there's no reason why it shouldn't work.

hughbe:
Fixed, thansk

PR #244, opened 2018-03-27
Fix font behaviour on Windows

hughbe:
- Font files are binary so don't corrupt them in fopen
- The type of WCHAR * is guint32, not wchar_t on Windows libgdiplus.

### DIFF
--- a/src/win32-private.c
+++ b/src/win32-private.c
@@ -40,7 +40,7 @@ FILE *CreateTempFile (char *filename)
 		return NULL;
 	}
 
-	return fopen (filename, "w");
+	return fopen (filename, "wb");
 }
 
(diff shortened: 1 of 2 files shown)


src/win32-private.c
```diff
@@ -40,7 +40,7 @@ FILE *CreateTempFile (char *filename)
 		return NULL;
 	}
 
-	return fopen (filename, "w");
+	return fopen (filename, "wb");
```
akoeplinger: there's also this in font.c which I assume should be fixed too:
https://github.com/mono/libgdiplus/blob/d9daa4c86b115f16b5e0a91f912a13baf5d1bbd5/src/font.c#L173

PR #251, opened 2018-03-29
Refactor region to remove RegionTypeRectF

hughbe:


src/region-private.h
```diff
@@ -45,18 +45,17 @@
 #define REGION_INFINITE_LENGTH		8388608
 
 typedef enum {
-        RegionTypeEmpty = 0,
-        RegionTypeRect  = 1,
-        RegionTypeRectF = 2,
-        RegionTypePath  = 3
+    RegionTypeRect     = 0x10000000,
```
hughbe: These now match the windows gdi+ values - which is necessary for compatibility with region data

Issue #237, opened 2018-03-27
Pango text rendering is broken on macOS


filipnavara:
The Pango code (builds with `--with-pango`) doesn't work on macOS. Pango defaults to the CoreText rendering and mapping backend, which libgdiplus doesn't support. Workaround is to set `PANGOCAIRO_BACKEND=fc`.

In order to support the CoreText backend the following is minimum necessary changes:

- [ ] Add support for CoreText fonts to gdip_get_fontfamily_details
- [ ] Add CoreText font registration to GdipPrivateAddFontFile and GdipPrivateAddMemoryFont
- [ ] Workaround/fix bug with font sizes in Pango (Pango incorrectly treats CoreGraphics points as typographic points and screws up the sizes by factor of 96f / 72f on displays)

Alternatively, we can set `PANGOCAIRO_BACKEND` ourselves until the above issues are fixed.

filipnavara:
I have local patches to implement first two items on the list. The last one is a bit more tricky and proper fix would be to fix Pango.

filipnavara:
I decided to abandon the CoreText branch for now and get at least the FontConfig code working properly. In case someone decides to revive it here's the link to unfinished code: https://github.com/filipnavara/libgdiplus/commit/7b6322bb6d10d1eaa10701eafda68675d89d5e71

PR #258, opened 2018-03-29
Use our own pango font maps to support custom fonts and use FontConfig backend on macOS by default

filipnavara:
Fixes issue #237.

Pango has three backends - CoreText, Win32 and FontConfig. By default it tries to select the one that is most appropriate for the platform. However the current libgdiplus code depends on FontConfig and FreeType directly, so we need to force the use of FontConfig backend. Additionally it fixes an issue with custom fonts, which use custom FcConfig objects that were not properly propagated to Pango. Lastly, it fixes a memory leak in gdip_get_fontfamily_details.

### DIFF
--- a/src/font-private.h
+++ b/src/font-private.h
@@ -39,6 +39,7 @@
 
 #ifdef USE_PANGO_RENDERING
 	#include <pango/pangofc-font.h>
+	#include <pango/pangofc-fontmap.h>
 	#include <pango/pangocairo.h>
 #endif
 
@@ -58,6 +59,7 @@ struct _Font {
 };
 
 void gdip_font_clear_pattern_cache (void) GDIP_INTERNAL;
+void gdip_delete_system_fonts (void) GDIP_INTERNAL;
 
 #ifdef USE_PANGO_RENDERING
 
--- a/src/font.c
+++ b/src/font.c
@@ -74,6 +74,12 @@ gdip_fontfamily_new ()
 
 static GpFontCollection *system_fonts = NULL;
 
+void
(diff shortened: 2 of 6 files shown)


kasper3:
Why is `USE_PANGO_RENDERING` an optional thing? Can this be made required dependency for libgdiplus package? Since this library has general-purpose use and there is no way for downstream to tell what features were supported at compile time, it might make sense to avoid surprises like https://github.com/dotnet/corefx/issues/28352. Like normally if I am not using the feature of certain dependency, the shared library does not get loaded until required:

* For build machine libpango will be required.
* At runtime, the target OS may or may not have libpango. It would be a runtime error if user starts using a pango feature.

Even in the current situation, if we build with pango and target system doesn't have libpango, then it works until we hit the code path leading to pango usage.

Keeping it optional for build machine machine is an unwarranted territory which may then require some kind of a query API for downstream (corefx) to check the capability?

filipnavara:
@kasper3 In principle I agree that `USE_PANGO_RENDERING` should be used whenever Pango library is available on the system, which should be true for most Linux systems. However, that's not my call to make since I am merely a contributor, not a maintainer.

For a long time the option was marked as unsupported. I spent the past week resurrecting it and making sure it behaves correct enough to run basic System.Windows.Forms applications. I'm now looking into making unit tests to ensure that all the edge case behavior is correct.

Once the code is covered by tests we should make the option supported. Perhaps even kill the `--with-pango` option and just test for its presence in the build script by default. I am open to opinions and feedback from package maintainers on this.

PR #262, opened 2018-03-30
RFC: Add Wine tests for GdipDrawString, GdipMeasureString and GdipMeasureCharacterRanges

filipnavara:
This is an experiment with porting over some unit tests from Wine. There's a lot of untested scenarios with the text functions and this seems to be the fastest way to get the code under unit tests so we can start fixing it.

The formatting is kept from Wine, not sure if it's worth changing. The test macros are defined on top of the file, but if there's a demand to port over more of the Wine tests then maybe we should move them to testhelpers.h along with the helper functions.


filipnavara:
I accidently run the tests with the Pango backend instead of the Cairo one and thus some of them are not correctly prefixed with TODO and fail.

filipnavara:
Turns out that the LICENSE may prevent us from taking the code from Wine. The Wine code is LGPL, while the libgdiplus is dual-licensed under LGPL and MPL.

PR #263, opened 2018-03-30
Run System.Drawing corefx tests in CI

akoeplinger:
/cc @filipnavara


filipnavara:
It works! Wunderbar :-)

PR #266, opened 2018-03-31
Restore EMF pen behaviour and increase compatability with WMF pen behaviour

hughbe:
@filipnavara

### DIFF
--- a/src/metafile.c
+++ b/src/metafile.c
@@ -495,6 +495,24 @@ gdip_metafile_CreatePenIndirect (MetafilePlayContext *context, DWORD style, DWOR
 	GdipSetPenStartCap (pen, line_cap);
 	GdipSetPenEndCap (pen, line_cap);
 
+	
+	s = (style & PS_JOIN_MASK);
+	switch (s) {
+	default:
+		g_warning ("Invalid pen join, style %d, (style & PS_JOIN_MASK) %d", style, s);
+		/* fall through */
+	case PS_JOIN_ROUND:
+		line_join = LineJoinRound;
+		break;
+	case PS_JOIN_BEVEL:
+		line_join = LineJoinBevel;
+		break;
+	case PS_JOIN_MITER:
+		line_join = LineJoinMiter;
+		break;
+	}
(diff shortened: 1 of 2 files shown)


src/metafile.c
```diff
@@ -494,20 +541,29 @@ gdip_metafile_CreatePenIndirect (MetafilePlayContext *context, DWORD style, DWOR
...
+		line_join = LineJoinRound;
+		break;
+	case PS_JOIN_BEVEL:
+		line_join = LineJoinBevel;
+		break;
+	case PS_JOIN_MITER:
+		line_join = LineJoinMiter;
+		break;
+	}
+	GdipSetPenLineJoin (pen, line_join);
```
filipnavara: I still thing this block should be wrapped in `if (dwPenStyle & PS_GEOMETRIC)`.

filipnavara: ...including the line caps (https://msdn.microsoft.com/en-us/library/windows/desktop/dd162711(v=vs.85).aspx).

hughbe:
> In terms of avoiding code duplication maybe gdip_metafile_CreatePenIndirect can be implemented by calling gdip_metafile_ExtCreatePen?

Sort of, but then there would be subtly different behaviour to Windows GDI+. If you passed PS_GEOMETRIC to `CreatePen` nothing would happen currently (and this matches Windows GDI+), but if we implemented `createPen` in terms of `extCreatePen`, then the PS_GEOMETRIC flag would be listened to instead of ignored.

filipnavara:
You cannot pass PS_GEOMETRIC to gdip_metafile_CreatePenIndirect in WMF because the values are read from 16-bit WORD and thus it can never be present in the data.

filipnavara:
There are no test for EMR_CREATEPEN in EMF, but I expect it to respect the PS_GEOMETRIC style. Even if it was not, then it's simple to just mask it out in gdip_metafile_CreatePenIndirect.

hughbe:
Derp you're absolutely right that's super clever! Fixed - good to have your eyes on this :D

PR #267, opened 2018-03-31
Override inc language classification

kasper3:
`alpha-premul-table.inc` is classified as PHP, that amounts to 13.4% of the code:

image

With this override, linguist will classify the inc file as C.

Related issue https://github.com/github/linguist/issues/4082.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,2 @@
 *.sh	crlf=input
-
+*.inc	linguist-language=C


kasper3:
@akoeplinger, I used the first method of overriding: https://github.com/github/linguist/blob/dc328761137b14b1273baaad56d53f89f66af88e/README.md#overrides

PR #268, opened 2018-03-31
Add initial tests for GdipMeasureString and fix Pango text backend to pass them

filipnavara:
Unfortunately the Cairo text backend is too broken for the tests to be enabled there, so I initially enable them only when Pango text backend is used. Even now some of the tests are disabled because of small issues in the Pango code.

### DIFF
--- a/src/text-pango.c
+++ b/src/text-pango.c
@@ -486,7 +486,7 @@ gdip_pango_setup_layout (GpGraphics *graphics, GDIPCONST WCHAR *stringUnicode, i
 	case StringAlignmentNear:
 		break;
 	case StringAlignmentCenter:
-		box_offset->Y += (FrameHeight - box->Height) * .5;
+		box_offset->Y += (FrameHeight - box->Height) * 0.5;
 		break;
 	case StringAlignmentFar:
 		box_offset->Y += (FrameHeight - box->Height);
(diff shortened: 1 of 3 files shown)


tests/testtext.c
```diff
@@ -0,0 +1,214 @@
...
+	status = GdipMeasureString (graphics, teststring1, 3, font, &rect, format, &bounds, &glyphs, &lines);
+	expect (Ok, status);
+	TODO expect (3, glyphs);
+	expect (2, lines);
+
+	// ...and lastly check the correct results for adding another character past the last new line
+	set_rect_empty (&bounds);
+	status = GdipMeasureString (graphics, teststring1, 4, font, &rect, format, &bounds, &glyphs, &lines);
+	expect (Ok, status);
+	TODO expect (4, glyphs);
```
hughbe: Do we still need these todos?

hughbe: (I’m assuming that these are for wine instead of for libgdiplus)

Maybe we can also use the #if USE_WINDOWS_GDIPLUS macro to be clearer about this

filipnavara: These test are not from Wine (unlike the other PR). Yes, we need them, unfortunately.

filipnavara: I can wrap the `TODO` macro with `#if USE_WINDOWS_GDIPLUS`. I'd prefer that over having `#if` all over the place.

filipnavara:
@monojenkins build

src/text-pango.c
```diff
@@ -481,25 +486,25 @@ gdip_pango_setup_layout (GpGraphics *graphics, GDIPCONST WCHAR *stringUnicode, i
 	case StringAlignmentNear:
 		break;
 	case StringAlignmentCenter:
-		box_offset->Y += (rc->Height - box->Height) / 2;
+		box_offset->Y += (FrameHeight - box->Height) * .5;
```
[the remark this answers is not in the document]
filipnavara: Done.

PR #273, opened 2018-04-01
Fix GdipCreateBitmapFromGraphics to actually use the graphics

hughbe:
And simplify it at the same time to use GdipCreateBitmapFromScan0

### DIFF
--- a/src/bitmap.c
+++ b/src/bitmap.c
@@ -29,6 +29,7 @@
 
 #include "gdiplus-private.h"
 #include "bitmap-private.h"
+#include "graphics-private.h"
 
 
 static GpStatus gdip_bitmap_clone_data_rect (BitmapData *srcData, Rect *srcRect, BitmapData *destData, Rect *destRect);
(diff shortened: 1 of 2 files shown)


src/bitmap.c
```diff
@@ -973,49 +974,21 @@ GdipCreateBitmapFromScan0 (int width, int height, int stride, PixelFormat format
...
+	if (!graphics || !bitmap)
+		return InvalidParameter;
 
-	bitmap_data = gdip_frame_add_bitmapdata(frame);
-	if (bitmap_data == NULL) {
-		goto fail;
-	}
+	status = GdipCreateBitmapFromScan0 (width, height, 0, PixelFormat32bppPARGB, NULL, &result);
+	if (status != Ok)
+		return status;
```
filipnavara: Shouldn't it use the pixel format of the passed `graphics`?

hughbe: I don't think so according to my testing. It always creates a PixelFormat32PARGB image

filipnavara: Okay, I wasn't sure. Wine tests also imply PixelFormat32PARGB is used.

PR #274, opened 2018-04-01
Allow redefinition of WINGDIPAPI

filipnavara:


filipnavara:
@hughbe Any idea how I managed to break the MSVC build so bad with this? Apparently it doesn't like the changes in general.h, which is weird, because my VS2017 build tools didn't like it the other way around.

hughbe:
Could it be something to do with the fact that `WINGDIPAPI` is defined in `gdiplus-private`?

#ifdef _MSC_VER
#define WINGDIPAPI __declspec( dllexport )
#else
#define WINGDIPAPI
#endif

hughbe:
Could it be something to do with the fact that `WINGDIPAPI` is defined in `gdiplus-private`?

#ifdef _MSC_VER
#define WINGDIPAPI __declspec( dllexport )
#else
#define WINGDIPAPI
#endif

filipnavara:
My idea was to use the same set of headers for building tests on Windows and on Linux. I had problems with the C++ references on some parameters expressed as (GpRectF &) in the Windows headers. On the other side the GdiPlusFlat.h in the repository doesn't specify the correct calling convention (__stdcall) and this was an attempt to fix that - by allowing `-D WINGDIPAPI=__stdcall` in the project file.

I think it would make sense to clean it up and also move all the `#include` code to testhelpers.h. Anyway, closing this PR for now, will fix it up later unless someone else beats me to it.

Issue #189, opened 2017-12-17
16bpp pixel values do not match GDI+


hughbe:
In `test_validImage16bppBitmapV3Header` in `testbmpcodec.c`

```cpp
static void test_validImage16bppBitmapV3Header ()
{
	BYTE image1x1RGB565[] = {
		// -- BITMAPCOREHEADER -- //
		/* Signature */ 0x42, 0x4D,
		/* File Size */ 0x4A, 0x00, 0x00, 0x00,
		/* Reserved */  0x00, 0x00, 0x00, 0x00,
		/* Offset */    0x46, 0x00, 0x00, 0x00,
		// -- BITMAPINFOHEADER
		/* Header Size */      0x38, 0x00, 0x00, 0x00,
		/* Width */            0x01, 0x00, 0x00, 0x00,
		/* Height */           0x01, 0x00, 0x00, 0x00,
		/* Planes */           0x01, 0x00,
		/* Bit Count */        0x10, 0x00,
		/* Compression */      0x00, 0x00, 0x00, 0x00,
		/* Image Size */       0x00, 0x00, 0x00, 0x00,
		/* Horizontal */       0x00, 0x00, 0x00, 0x00,
		/* Vertical */         0x00, 0x00, 0x00, 0x00,
		/* Colors Used */      0x00, 0x00, 0x00, 0x00,
		/* Important Colors */ 0x00, 0x00, 0x00, 0x00,
		// -- BITMAPV3INFOHEADER -- //
		/* Red Mask */         0x00, 0xF8, 0x00, 0x00,
		/* Green Mask */       0xE0, 0x07, 0x00, 0x00,
		/* Blue Mask */        0x1F, 0x00, 0x00, 0x00,
		/* Alpha Mask */       0x00, 0x00, 0x00, 0x00,
		// -- Image Data --/
		0xD7, 0xFE, 0x00, 0x00
	};
	BYTE image1x1RGB555[] = {
		// -- BITMAPCOREHEADER -- //
		/* Signature */ 0x42, 0x4D,
		/* File Size */ 0x4A, 0x00, 0x00, 0x00,
		/* Reserved */  0x00, 0x00, 0x00, 0x00,
		/* Offset */    0x46, 0x00, 0x00, 0x00,
		// -- BITMAPINFOHEADER
		/* Header Size */      0x38, 0x00, 0x00, 0x00,
		/* Width */            0x01, 0x00, 0x00, 0x00,
[...]

Issue #157, opened 2017-12-15
Support 16bpp BITMAPV4HEADER bitmaps


hughbe:
In `test_validImage16bppBitmapV4Header` in `testbmpcodec.c`

```cpp
// FIXME: this causes a stack buffer overflow (crash) in libgdiplus.
```

16bpp BITMAPV4HEADER cause libgdiplus to crash

PR #276, opened 2018-04-02
Fix bmpcodec 16bpp decoding

hughbe:
Fixes https://github.com/mono/libgdiplus/issues/189
Fixes https://github.com/mono/libgdiplus/issues/157

### DIFF
--- a/src/bmpcodec.c
+++ b/src/bmpcodec.c
@@ -84,8 +84,13 @@ gdip_get_bmp_pixelformat (BITMAPV5HEADER *bih, PixelFormat *dest)
 	case BI_BITFIELDS:
 		if (bitCount != 16)
 			return OutOfMemory;
-		/* note: incomplete at this stage */
-		*dest = PixelFormat16bppRGB565;
+
+		if ((bih->bV5RedMask == 0x7C00) && (bih->bV5GreenMask == 0x3E0) && (bih->bV5BlueMask == 0x1F))
+			*dest = PixelFormat16bppRGB555;
+		else if ((bih->bV5RedMask == 0xF800) && (bih->bV5GreenMask == 0x7E0) && (bih->bV5BlueMask == 0x1F))
+			*dest = PixelFormat16bppRGB565;
+		else
+			*dest = PixelFormat32bppRGB;
 		break;
 	default:
 		switch (bitCount) {
@@ -99,7 +104,7 @@ gdip_get_bmp_pixelformat (BITMAPV5HEADER *bih, PixelFormat *dest)
 			*dest = PixelFormat24bppRGB;
 			break;
 		case 16:
-			*dest = PixelFormat16bppRGB565;
+			*dest = PixelFormat16bppRGB555;
 			break;
 		case 8:
 			*dest = PixelFormat8bppIndexed;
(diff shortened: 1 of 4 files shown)


src/bmpcodec.c
```diff
@@ -827,6 +839,34 @@ gdip_read_BITMAPINFOHEADER (void *pointer, ImageSource source, BITMAPV5HEADER *b
...
+	bmi->bV5GreenMask = (data_read[3]<<24 | data_read[2]<<16 | data_read[1]<<8 | data_read[0]);
+
+	dw = 0;
+	size_read = gdip_read_bmp_data (pointer, data_read, size, source);
+	if (size_read < size)
+		return OutOfMemory;
+	bmi->bV5BlueMask = (data_read[3]<<24 | data_read[2]<<16 | data_read[1]<<8 | data_read[0]);
+
+	return Ok;
 }
```
filipnavara: This doesn't sound right. I don't think the masks are even present there with BITMAPINFOHEADER when `biCompression` is `BI_RGB`. In that case it should use the default 16 bits per pixel values:

```
bmi->red_mask = 0x7C00;
bmi->green_mask = 0x3E0;
bmi->blue_mask = 0x1F;
```

tests/testtext
```diff
@@ -0,0 +1,210 @@
+#! /bin/sh
```
filipnavara: This is now .gitignored right? Shouldn't be part of the commit. :-)

src/bmpcodec.c
hughbe: The cause of this issue was https://github.com/mono/libgdiplus/issues/285 which needs to be fixed (I'll do so separately)

Addressed this feedback though

src/bmpcodec.c
```diff
@@ -738,7 +750,7 @@ gdip_read_BITMAPINFOHEADER (void *pointer, ImageSource source, BITMAPV5HEADER *b
 		bmi->bV5ClrUsed = 0;
 		bmi->bV5ClrImportant = 0;
 
-		return Ok;
+		goto fill_rgb_masks;
```
filipnavara: This looks kind of useless. There will never be masks since there's no compression and hence no BI_BITFIELDS (`bV5Compression` should be set to `BI_RGB` for clarity).  And it will never be 16-bit since the allowed values for bcBitCount are only 1, 4, 8, and 24 (https://msdn.microsoft.com/en-us/library/windows/desktop/dd183372(v=vs.85).aspx).

src/bmpcodec.c
```diff
@@ -827,6 +839,41 @@ gdip_read_BITMAPINFOHEADER (void *pointer, ImageSource source, BITMAPV5HEADER *b
...
+		bmi->bV5BlueMask = 0;
+		return Ok;
+	}
+	if (bmi->bV5Compression != BI_BITFIELDS) {
+		// Non-compressed 16bpp bitmaps are always RGB555.
+		bmi->bV5RedMask = 0x7C00;
+		bmi->bV5GreenMask = 0x3E0;
+		bmi->bV5BlueMask = 0x1F;
+		return Ok;
+	}
```
filipnavara: These two block could probably be moved up to the "goto fill_rgb_masks" point and then the mask reading code could be shared instead of being duplicated. In fact, I think the `if (bmi->bV5BitCount != 16)` block is unnecessary.

src/bmpcodec.c
```diff
@@ -784,8 +796,29 @@ gdip_read_BITMAPINFOHEADER (void *pointer, ImageSource source, BITMAPV5HEADER *b
...
+				return OutOfMemory;
+			bmi->bV5GreenMask = (data_read[3]<<24 | data_read[2]<<16 | data_read[1]<<8 | data_read[0]);
+
+			dw = 0;
+			size_read = gdip_read_bmp_data (pointer, data_read, size, source);
+			if (size_read < size)
+				return OutOfMemory;
+			bmi->bV5BlueMask = (data_read[3]<<24 | data_read[2]<<16 | data_read[1]<<8 | data_read[0]);
+		}
+
```
filipnavara: Any reason why you don't fall through to the code below? Also, now the implicit BI_RGB masks from previous change are missing...

filipnavara: For the record I don't mind the code duplication if it makes the code more readable. It's matter of personal opinion and I find your version little more readable even though it's a bit verbose... but the BI_RGB part has to be fixed in either case.

PR #277, opened 2018-04-02
Convert Windows to use vckpg instead of nuget

hughbe:
This is better as we don't need to maintain our own dependency on hexchat's GTK and we can rely on the vcpkg community

@qmfrederik FYI


hughbe:
@filipnavara this should fix the WIn32 issue you mentioned in https://github.com/mono/libgdiplus/pull/269

filipnavara:
There was already Pango in the HexChat package, so it was not a deal-breaker, but I definitely prefer the VCPKG approach.

.jenkins-windows.sh
```diff
@@ -2,16 +2,16 @@
 
 export PATH=/usr/bin:/usr/local/bin:$PATH
 
-git clone https://github.com/Microsoft/vcpkg.git
-bootstrap-vcpkg.bat
-vcpkg integrate install
+git clone https://github.com/Microsoft/vcpkg.git /cygdrive/c/Program\ Files\ \(x86\)/
+/cygdrive/c/Program\ Files\ \(x86\)/bootstrap-vcpkg.bat
+/cygdrive/c/Program\ Files\ \(x86\)/vcpkg integrate install
```
filipnavara: This is not writable directory, so it's not going to work even if you get all the path quoting right.

filipnavara: You probably want something more like

```
git clone https://github.com/Microsoft/vcpkg.git
vcpkg/bootstrap-vcpkg.bat
vcpkg/vcpkg install giflib libjpeg-turbo libpng cairo glib tiff libexif glib pango
```

hughbe:
@akoeplinger does this require vcpkg to be installed on the mono CI? I'm not sure whether I can do it from the script:

> ./.jenkins-windows.sh: line 6: vcpkg/bootstrap-vcpkg.bat: Permission denied

qmfrederik:
@hughbe Yes, vcpkg is the way to go! The build scripts look a lot nicer now 😄 .

src/libgdiplus.vcxproj
```diff
@@ -87,28 +87,17 @@
   <PropertyGroup Label="UserMacros" />
   <ItemDefinitionGroup>
     <ClCompile>
-      <AdditionalIncludeDirectories>$(ProjectDir)\..;$(ProjectDir)\..\gtk\$(Platform)\release\include;$(ProjectDir)\..\gtk\$(Platform)\release\include\glib-2.0;$(ProjectDir)\..\gtk\$(Platform)\release\lib\glib-2.0\include;$(ProjectDir)\..\gtk\$(Platform)\release\include\cairo;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
-      <PreprocessorDefinitions>HAVE_LIBGIF;HAVE_LIBJPEG;HAVE_LIBTIFF;HAVE_LIBPNG;HAVE_FCFINI;_WINDLL;WIN32;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions>HAVE_LIBGIF;HAVE_LIBJPEG;HAVE_LIBTIFF;HAVE_LIBPNG;_WINDLL;WIN32;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
filipnavara: Where did HAVE_FCFINI go, btw?

ras0219-msft:
We'd also be happy to accept a PR in vcpkg for `libgdiplus` so we can make sure dependency updates don't break you (or at a minimum notify you) :)

akoeplinger:
@hughbe hm, I don't know why it fails that way. If I run the command manually as the same user via RDP it works. I'll need to dig in why :)

filipnavara:
@monojenkins build

filipnavara:
I am not surprised it fails. The "vcpkg integrate install" command install system-wide MSBuild scripts.

hughbe:
Maybe we need to call it just once on the machine (not from the script) to install it?

akoeplinger:
@filipnavara it doesn't fail at `vcpkg integrate install`, it fails right away for some reason.

@hughbe we could do that but since the Windows builders are dynamically provisioned Azure VMs we'd need to do it at provision time. And since vcpkg compiles all the packages that takes quite some time so we'll probably need to use the caching feature instead.

akoeplinger:
Ok, the issue was some weird interaction with the cygwin bash shell calling the Windows batch command.

Moving everything to powershell seemed to work. Build without caching (i.e. compiling all the libs) takes about 30mins now which is OK for now I think, the Travis build takes longer anyway.

Thanks!

PR #278, opened 2018-04-02
Implement GdipGetImageThumbnail

hughbe:


src/image.c
```diff
@@ -2651,11 +2651,54 @@ GdipGetEncoderParameterList (GpImage *image, GDIPCONST CLSID *clsidEncoder, UINT
...
-	/* We don't really need to implement this; we've got it handled in System.Drawing directly */
-	return NotImplemented;
+	if ((thumbWidth && !thumbHeight) || (!thumbWidth && thumbHeight))
+		return OutOfMemory;
+
+	if (!thumbWidth || !thumbHeight)
+		thumbWidth = thumbHeight = 120;
+
+	if (!thumbHeight)
+		thumbHeight = 120;
```
filipnavara: This is already handled by the previous `if`.

filipnavara: Maybe something like this would be easier to read?

```
if (!thumbWidth && !thumbHeight) {
  thumbWidth = thumbHeight = 120;
} else if (!thumbWidth || !thumbHeight) {
  return OutOfMemory;
}
```

PR #287, opened 2018-04-06
Fix GetRegionBounds for large path regions

hughbe:
And add tests for GdipCreateRegionPath

Fixes https://github.com/mono/libgdiplus/issues/304

### DIFF
--- a/src/region.c
+++ b/src/region.c
@@ -1493,6 +1493,10 @@ GdipGetRegionBounds (GpRegion *region, GpGraphics *graphics, GpRectF *rect)
 	if (region->type == RegionTypePath) {
 		GpRect bounds;
 
+		/* optimisation for simple path */
+		if (region->tree->path)
+			return GdipGetPathWorldBounds (region->tree->path, rect, NULL, NULL);
+
 		gdip_region_bitmap_ensure (region);
 		if (!region->bitmap)
 			return OutOfMemory;
--- a/tests/testregion.c
+++ b/tests/testregion.c
@@ -191,9 +191,11 @@ static void test_createRegionPath ()
 
 	status = GdipCreateRegionPath (path, &region);
 	assertEqualInt (status, Ok);
-	// FIXME: GetRegionBounds fails with OutOfMemory.
+	// FIXME: libgdiplus does not identify this as an infinite region.
 #if defined(USE_WINDOWS_GDIPLUS)
 	verifyRegion (region, -4194304.0f, -4194304.0f, 8388608.0f, 8388608.0f, FALSE, TRUE);
+#else
+	verifyRegion (region, -4194304.0f, -4194304.0f, 8388608.0f, 8388608.0f, FALSE, FALSE);
 #endif
 	GdipDeleteRegion (region);
 	GdipDeletePath (path);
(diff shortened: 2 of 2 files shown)


src/region-bitmap.c
```diff
@@ -537,9 +537,7 @@ gdip_region_bitmap_from_path (GpPath *path)
 
 	/* here the memory is allocated for a ARGB bitmap - so 32 times bigger than our normal alpha bitmap */
 	if ((size < 1) || (size > REGION_MAX_BITMAP_SIZE << 5)) {
-		g_warning ("Path conversion requested %lu bytes (%d x %d). Maximum size is %d bytes.",
-			size, bounds.Width, bounds.Height, REGION_MAX_BITMAP_SIZE << 5);
-		return NULL;
+		return alloc_bitmap_with_buffer (REGION_INFINITE_POSITION, REGION_INFINITE_POSITION, REGION_INFINITE_LENGTH, REGION_INFINITE_LENGTH, NULL);
```
filipnavara: Not sure about this. How is it handled by `alloc_merged_bitmap` and all the region combine operations? Are we sure it is never reached?

src/region.c
```diff
@@ -182,7 +182,7 @@ gdip_is_InfiniteRegion (GpRegion *region)
 		/* FIXME: incomplete and not 100% accurate (curves) - but cover the most common case */
 		if (region->tree && region->tree->path && (region->tree->path->count == 4)) {
 			GpRectF bounds;
-			if (GdipGetPathWorldBounds (region->tree->path, &bounds, NULL, NULL) != Ok)
+			if (GdipGetPathWorldBounds (region->tree->path, &bounds, NULL, NULL) == Ok)
```
filipnavara: Weirdly enough, this seems to trip one of the unit tests (creating a path, adding a curve with REGION_INFINITE_POSITION/REGION_INFINITE_LENGTH coordinates and checking if a region created from the path is infinite).

hughbe:
Okay I've updated this PR only to include unit tests as it looks like there's a lot of behaviour I don't understand here...

filipnavara:
@hughbe  Does this work?

```csharp
GpStatus WINGDIPAPI
GdipGetRegionBounds (GpRegion *region, GpGraphics *graphics, GpRectF *rect)
{
	if (!region || !graphics || !rect)
		return InvalidParameter;

	if (region->type == RegionTypePath) {
		GpRect bounds;

		if (region->tree->path)
			return GdipGetPathWorldBounds (region->tree->path, rect, NULL, NULL);

		gdip_region_bitmap_ensure (region);
		if (!region->bitmap)
			return OutOfMemory;

		/* base the bounds on the reduced bitmap of the paths */
		gdip_region_bitmap_get_smallest_rect (region->bitmap, &bounds);

		/* small loss of precision when converting the bitmap coord to float */
		rect->X = bounds.X;
		rect->Y = bounds.Y;
		rect->Width = bounds.Width;
		rect->Height = bounds.Height;
	} else {
		gdip_get_bounds (region->rects , region->cnt, rect);
	}

	return Ok;
}

```

hughbe:
Looks like this broke some tests - maybe a precision issue? Could it be a bug in GdipGetPathWorldBounds?

filipnavara:
@hughbe Yep, definitely some issue with rounding error. I will fix it once #227 is merged. There's no need to do path flattening in `GdipGetPathWorldBounds` and it should make it faster and avoid possible rounding issues.

filipnavara:
Hmm, actually the rounding error may be introduce earlier. I will replicate it on my machine.

hughbe:
Closing in favor  of #311

PR #314, opened 2018-04-09
Fix CGRect and CGSize definitions on 64-bit systems.

filipnavara:


src/graphics-private.h
```diff
@@ -169,19 +169,26 @@ void gdip_cairo_curve_to (GpGraphics *graphics, double x1, double y1, double x2,
 	BOOL convert_units, BOOL antialiasing) GDIP_INTERNAL;
 
 #ifdef CAIRO_HAS_QUARTZ_SURFACE
+
+#if __x86_64__
+typedef double nfloat;
```
hughbe: Would it make more sense to call this `CGFloat` to match cocoa?

filipnavara: Yes, it would. Good point :-)

Issue #195, opened 2017-12-19
GdipFillRegion produces an incorrect result in excluded area


emptyVoid:
## Steps to Reproduce

```c++
#ifdef _WIN32
#   include <Windows.h>
#   include <gdiplus.h>
    using namespace Gdiplus;
    using namespace Gdiplus::DllExports;
#   pragma comment(lib, "gdiplus.lib")
#endif

#include <GdiPlusFlat.h>
#include <string>

#ifdef _WIN32
#   define WIDEN( x ) L##x
using ucs2_string = std::wstring;
#else
#   define WIDEN( x ) u##x
using ucs2_string = std::u16string;
#endif

bool get_encoder_clsid( const ucs2_string& format, CLSID* clsid );

int main()
{
    ULONG_PTR           token = 0;
    GdiplusStartupInput input = { 0 };
    input.GdiplusVersion      = 1;
    GdiplusStartup( &token, &input, nullptr );

    GpBitmap* bitmap = nullptr;
    GdipCreateBitmapFromScan0( 400, 400, 0, PixelFormat32bppARGB, nullptr, &bitmap );

    GpGraphics* graphics = nullptr;
    GdipGetImageGraphicsContext( bitmap, &graphics );

    GdipGraphicsClear( graphics, 0xFF808080 );

    GpPath* rectPath = nullptr;
    GdipCreatePath( FillModeAlternate, &rectPath );
    GdipAddPathRectangleI( rectPath, 50, 50, 300, 300 );

    GpPath* polyPath = nullptr;
    GdipCreatePath( FillModeAlternate, &polyPath );
    GpPoint polyPoints[] = { { 100, 100 },
                             { 200,  75 },
                             { 300, 100 },
                             { 325, 200 },
                             { 300, 300 },
                             { 200, 325 },
                             { 100, 300 },
                             {  75, 200 } };
    GdipAddPathPolygonI( polyPath, polyPoints, sizeof( polyPoints ) / sizeof( polyPoints[0] ) );

    GpRegion* region = nullptr;
    GdipCreateRegion( &region );
    GdipSetEmpty( region );
    GdipCombineRegionPath( region, rectPath, CombineModeUnion );
    GdipCombineRegionPath( region, polyPath, CombineModeExclude );

    GpSolidFill* brush = nullptr;
    GdipCreateSolidFill( 0xFF00FF00, &brush );

    GdipFillRegion( graphics, brush, region );

    GdipDeleteGraphics( graphics );
[...]

PR #317, opened 2018-04-10
Use native cairo functions for rendering region mask bitmaps.

filipnavara:
This is alternative patch to #283 for issue #195. It reduces memory allocations and moves the hard part of rendering to Cairo.


hughbe:
Should this still be opened considering its billed as an alternative to #283 which has been merged

filipnavara:
Yes, it should be rebased and merged. :)

hughbe:
@marek-safar looks like this one and #328 are good to merge then :)

PR #326, opened 2018-04-11
Fix various region scan behaviour

hughbe:
Fixes #290
Fixes #289

~Depends on #311~
~Depends on #286~
~Depends on #293~

PR #286, opened 2018-04-06
Fix GdipIsRegionEmpty for negative rects

hughbe:
Found when running the region tests against Windows GDI+

PR #327, opened 2018-04-11
Fix various region visibility edge cases

hughbe:
Fixes part of #291

~Depends on #326~
~Depends on #311~
~Depends on #286~
~Depends on #293~

### DIFF
--- a/src/region.c
+++ b/src/region.c
@@ -101,6 +101,29 @@ gdip_is_Point_in_RectFs_Visible (float x, float y, GpRectF* r, int cnt)
 	return FALSE;
 }
 
+static BOOL
+gdip_is_Rect_in_RectF_Visible (float x, float y, float width, float height, GpRectF* rect)
+{
+	if (rect->Width == 0 || rect->Height == 0)
+		return FALSE;
+
+	return x < rect->X + rect->Width && x + width > rect->X && y < rect->Y + rect->Height && y + height > rect->Y;
+}
+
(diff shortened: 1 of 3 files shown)


src/region.c
```diff
@@ -101,6 +101,29 @@ gdip_is_Point_in_RectFs_Visible (float x, float y, GpRectF* r, int cnt)
 	return FALSE;
 }
 
+static BOOL
+gdip_is_Rect_in_RectF_Visible (float x, float y, float width, float height, GpRectF* rect)
+{
+	if (rect->Width == 0 || rect->Height == 0)
+		return FALSE;
+
+	return x < rect->X + rect->Width && x + width > rect->X && y < rect->Y + rect->Height && y + height > rect->Y;
```
filipnavara: `x + width > rect->X` -> `x + width >= rect->X`
`y + height > rect->Y` -> `y + height >= rect->Y`

hughbe: Doing this change causes the tests to fail. It seems that the bounds are not inclusive on the right and bottom

hughbe:
Odd.. Looks like there's an issue with Linux that doesn't repro on mac:

```
21:40:10 Expected: 0
21:40:10 Actual:   1
21:40:10 lt-testregion: testregion.c:904: test_isVisibleRegionRectI: Assertion `(int)isVisible == (int)(0)' failed.
21:40:10 

```

PR #334, opened 2018-04-12
Cleanup graphics draw/fill functions to avoid code duplication

hughbe:
- We can implement `Gdip{Draw/Fill}*I` functions in terms of `Gdip{Draw/Fill}`
- We can implement some single functions in terms of multiple functions (e.g.`Gdip{Draw/Fill}Rectangle` in terms of `Gdip{Draw/Fille}Rectangles`

This will also make it a lot easier to record data when metafile recording is implemented, as EMF+ records are only for the float/array versions of GDI+ draw/fill functions

### DIFF
--- a/src/graphics.c
+++ b/src/graphics.c
@@ -250,8 +250,10 @@ GdipCreateFromHDC (HDC hdc, GpGraphics **graphics)
 	    w, h);
 			
 	*graphics = gdip_graphics_new (surface);
-	if (!*graphics)
+	if (!*graphics) {
+		cairo_surface_destroy (surface);
 		return OutOfMemory;
+	}
 
 	(*graphics)->dpi_x = (*graphics)->dpi_y = gdip_get_display_dpi ();
 	cairo_surface_destroy (surface);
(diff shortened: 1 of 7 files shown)


src/graphics.c
```diff
@@ -888,20 +877,13 @@ GdipDrawBeziers (GpGraphics *graphics, GpPen *pen, GDIPCONST GpPointF *points, I
...
-	case GraphicsBackEndCairo:
-		return cairo_DrawBeziersI (graphics, pen, points, count);
-	case GraphicsBackEndMetafile:
-		return metafile_DrawBeziersI (graphics, pen, points, count);
-	default:
-		return GenericError;
-	}
+	pointsF = convert_points (points, count);
+	return GdipDrawBeziers (graphics, pen, pointsF, count);
 }
```
filipnavara: Memory leak?

src/graphics.c
```diff
@@ -987,17 +943,12 @@ GdipDrawLines (GpGraphics *graphics, GpPen *pen, GDIPCONST GpPointF *points, INT
...
-	switch (graphics->backend) {
-	case GraphicsBackEndCairo:
-		return cairo_DrawLinesI (graphics, pen, points, count);
-	case GraphicsBackEndMetafile:
-		return metafile_DrawLinesI (graphics, pen, points, count);
-	default:
-		return GenericError;
-	}
+	pointsF = convert_points (points, count);
+	return GdipDrawLines (graphics, pen, pointsF, count);
```
filipnavara: Memory leak?

src/graphics.c
hughbe: Good point. this applies to all of these cases

filipnavara: Yep. The rest of the patch seems ok to me.

PR #360, opened 2018-04-14
Cleanup PathGradientBrush code and fix some compat bugs

hughbe:
- Handle memory allocation failures better
- Implement GdipCreatePathGradientI in terms of GdipCreatePathGradient
- Add more tests
- Fix ordering of parameter validation to match GDI+
- Fix writing to positions when the blend count is 1, to match GDI+

### DIFF
--- a/src/lineargradientbrush.c
+++ b/src/lineargradientbrush.c
@@ -92,9 +92,11 @@ gdip_linear_gradient_new (void)
 {
 	GpLineGradient *result = (GpLineGradient *) GdipAlloc (sizeof (GpLineGradient));
 
-	if (result && gdip_linear_gradient_init (result) != Ok) {
+	if (result) {
+		if (gdip_linear_gradient_init (result) == Ok)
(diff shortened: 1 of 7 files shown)


src/lineargradientbrush.c
```diff
@@ -92,9 +92,11 @@ gdip_linear_gradient_new (void)
...
 	GpLineGradient *result = (GpLineGradient *) GdipAlloc (sizeof (GpLineGradient));
 
-	if (result && gdip_linear_gradient_init (result) != Ok) {
+	if (result) {
+		if (gdip_linear_gradient_init (result) == Ok)
+			return result;
+
 		GdipFree (result);
-		return NULL;
 	}
```
filipnavara: This is not right, it will return freed object if `gdip_linear_gradient_init` fails. The previous code was correct.

hughbe: fixed

Issue #325, opened 2018-04-11
GdipAddPath string crashes with Pango backend


hughbe:
```
(process:82679): Pango-CRITICAL **: 12:53:11.183: pango_cairo_update_context: assertion 'cr != NULL' failed
FAIL testgraphicspath (exit status: 139)
```

hughbe:
Fixed in #322

PR #365, opened 2018-04-16
Fix incorrect cast in Pango code in GdipAddPathString.

filipnavara:


hughbe:
Fixes https://github.com/mono/libgdiplus/issues/325

filipnavara:
@hughbe This should hopefully pass the `dumpPath` test now.

src/graphics-path.c
```diff
@@ -1236,14 +1234,28 @@ GdipAddPathString (GpPath *path, GDIPCONST WCHAR *string, int length,
 	GpRectF box;
 	GpPointF box_offset;
 	PangoLayout* layout; 
+	GpStringFormat *string_format = (GpStringFormat *) format;
 
-	cairo_save (cr);
-	layout = gdip_pango_setup_layout ((GpGraphics *)cr, string, length, font, layoutRect, &box, &box_offset, format, NULL);
+	if (!(format->formatFlags & StringFormatFlagsNoClip)) {
```
hughbe: The format is optional so this will crash with a null pointer dereference if the format is not passed

src/graphics-path.c
```diff
@@ -1236,14 +1234,28 @@ GdipAddPathString (GpPath *path, GDIPCONST WCHAR *string, int length,
...
+			GdipDeleteFont (font);
+			GdipFree (utf8);
+			cairo_destroy (cr);
+			cairo_surface_destroy (cs);
+			return status;
+		}
+		string_format->formatFlags |= StringFormatFlagsNoClip;
+	}
+
+	layout = gdip_pango_setup_layout (cr, string, length, font, layoutRect, &box, &box_offset, format, NULL);
```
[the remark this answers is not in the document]
filipnavara: Eh, forgot to save the file :-)

src/graphics-path.c
filipnavara: Right... will fix.

hughbe:
@filipnavara is this good to be merged? If @marek-safar could merge given the all clear it would unblock one of my prs

filipnavara:
I quickly glanced over it and it seems ok. It definitely shouldn't make things worse.

marek-safar:
Could you add tests for the fix?

filipnavara:
I'll see if I can produce some meaningful test. The gist of the change was to remove an incorrect type cast (`(GpGraphics *)cr`) which caused the code to randomly crash.

filipnavara:
@marek-safar There's already unit test in PR #322 that covers it (not ready to be merged iirc). Merging the PRs will inevitably result in conflicts and I would prefer not to make the situation worse. PR #227 is the first one that introduces the graphicspath tests. Once that is merged I can extract the additional unit tests from PR #322 and add them.

filipnavara:
Closing in favor of #322.

PR #368, opened 2018-04-17
Add Clang sanitizer documentation to readme and fix asan bug in emfcodec

hughbe:


hughbe:
@akoeplinger @filipnavara could it be a good idea to run this on the Mac OS CI? It would prevent future regressions/exploitable bugs

filipnavara:
Sounds reasonable to me. The Travis macOS builds are currently the fastest ones, so it should not slow down the whole build process overall. The build tools there also seem reasonably up to date, unlike the Ubuntu ones.

Issue #372, opened 2018-04-22
Incorrect alpha blending when drawing an image with an alpha channel over another image


Marlamin:
libgdiplus version: 4.2-1+b1 (Debian stable) and the version shipped in [this](https://www.nuget.org/packages/runtime.linux-x64.CoreCompat.System.Drawing/1.0.0-beta009) (unable to tell). 

I'm not easily able to compile master so apologies if this is already resolved. 

I'm working on an application that compiles game maps out of several textures that heavily use alpha blending, it is using .NET Core's System.Drawing.Common which relies on libgdiplus (either the one on system or shipped with the application through the earlier linked NuGet package). The issue happens when I use Graphics.DrawImage to draw an image heavily utilizing the alpha channel over another image. The output appears pretty distorted wherever the alpha channel is used. 

I have set up a repo with sample images/code [**here**](https://github.com/Marlamin/AlphaBlendingRepro) that I also used to generate the following images (I cropped them for the issues sake, though). 

Expected output (generated on Windows):
![image](https://user-images.githubusercontent.com/50212/39098803-89bc000a-4670-11e8-80d6-9f8c6b5534cd.png)

Actual output (generated on my Debian machine): 
![image](https://user-images.githubusercontent.com/50212/39098800-83429806-4670-11e8-9a61-e29de40abb26.png)

If anyone is willing to test the application with master, that would be great. If it is already fixed or a case of PEBKAC, my apologies. If it is not I'm happy to help test any solutions.

filipnavara:
Thanks for the report. I will look into it shortly. It seems to be issue with handling ARGB images (as opposed to PARGB, which is premultiplied).

Marlamin:
Set up a new VM on my PC to compile the branch with the fix. Result is MUCH improved (no more distortion etc, yay) but not entirely the same yet (there's still a sort of dark border compared to the Windows version). 

![image](https://user-images.githubusercontent.com/50212/39129580-24f2aab6-470b-11e8-9857-647e163f5f4c.png)

I've updated the repro repo sample images with an actual map background/layer for better visibility of the issue, border stands out instead of merging with the terrain.

![image](https://user-images.githubusercontent.com/50212/39131896-dc651602-4710-11e8-8627-d1ae315ffd20.png)

[...]

PR #373, opened 2018-04-22
Fix condition for bitmap premultiplication.

filipnavara:
Fixes issue #372.


directhex:
Windows build failure seems unrelated, they're all doing it now

filipnavara:
@monojenkins build failed

hughbe:
Can you add some tests for this?

filipnavara:
Yeah, I should add the tests. There's a repro in the issue: https://github.com/Marlamin/AlphaBlendingRepro ... easy to check by human eye, little bit harder to check in a test (but certainly not impossible).

filipnavara:
I added unit test only for the original problem and verified that it fails before the fix and succeeds after. Then I noticed that other DrawImage* functions and code paths suffer from exactly the same problem so I fixed them too. There should ideally be unit tests for these places too, but I don't think they are essential.

hughbe:
Awesome! Thanks

filipnavara:
Hmm, apparently it fails some Mono unit tests now. I will look into it.

PR #383, opened 2018-06-18
GraphicsPath fixes

PreferLinux:
- When using the Pango text backend, pass an actual GpGraphics object.
- For GdipAddPathString, use an infinite Cairo Recording surface. I was having problems with longer strings clipping on the old 1x1 image surface.
- In GdipAddPathString, fix checking whether the next path point is a CAIRO_PATH_CLOSE_PATH.
- Also change append to correctly handle the second of two identically-positioned points closing the path.

### DIFF
--- a/src/graphics-path.c
+++ b/src/graphics-path.c
@@ -1247,6 +1247,12 @@ GdipAddPathString (GpPath *path, GDIPCONST WCHAR *string, int length,
 	
 	if (string_format != format)
 		GdipDeleteStringFormat (string_format);
+
+	// If our Cairo context had a current point before laying out the path, Pango will have moved us back there.
+	// We don't want that when we process the path below, so clear it if set.
+	if (cairo_has_current_point(cr))
+		cairo_new_sub_path(cr);
+
 	}
 #else
 	{
--- a/src/graphics-path.c
+++ b/src/graphics-path.c
@@ -1241,7 +1241,7 @@ GdipAddPathString (GpPath *path, GDIPCONST WCHAR *string, int length,
 	}
 
 	layout = gdip_pango_setup_layout (cr, string, length, font, layoutRect, &box, &box_offset, string_format, NULL);
-	cairo_move_to (cr, layoutRect->X + box_offset.X, layoutRect->Y + box_offset.X);
+	cairo_move_to (cr, layoutRect->X + box_offset.X, layoutRect->Y + box_offset.Y);
 	pango_cairo_layout_path (cr, layout);
 	g_object_unref (layout);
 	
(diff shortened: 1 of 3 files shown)


PreferLinux:
Looks like the Linux Travis CI build crashed while retrieving packages, before even starting the actual build...

filipnavara:
@monojenkins build

filipnavara:
Yeah, Travis configuration seems to be broken. I will see if I can fix it, but in the meantime the Jenkins Linux build should be enough to cover for it.

src/graphics-path.c
```diff
@@ -1236,13 +1239,32 @@ GdipAddPathString (GpPath *path, GDIPCONST WCHAR *string, int length,
...
 	cairo_save (cr);
-	layout = gdip_pango_setup_layout ((GpGraphics *)cr, string, length, font, layoutRect, &box, &box_offset, format, NULL);
+	
+	gr = (GpGraphics *) GdipAlloc(sizeof(GpGraphics));
+	gr->ct = cr;
+	// Also initialise anything needed by gdip_cairo_rectangle, called by gdip_pango_setup_layout.
+	gr->page_unit = UnitPixel;
+	gr->aa_offset_x = CAIRO_AA_OFFSET_X;
+	gr->aa_offset_y = CAIRO_AA_OFFSET_Y;
+
```
PreferLinux: The fix in #365 is definitely superior to this.

PreferLinux:
Rebased.

hughbe:
can you give an example of what this improves/fixes. Ideally we could use this as a chance to get some tests for these APIs in too

PreferLinux:
Well, there are three things dealt with (now that #322 has been merged, which fixed the other problem):
- In GdipAddPathString, use a Cairo Recording surface. For me, the existing Image surface will only show the first part of the string if the string is long enough (which is not very long, about 20 or so characters). A test would be reasonably easy.
- Similarly, don't include the point that Pango moves back to – if you draw a right-aligned string of width _w_ in a box with x = 0 and width = 2*_w_, the bounds of the graphics path will include 0 even though they obviously shouldn't. A test would be easy.
- Close sub-paths correctly. Some of this is in `append` (close the previous point if it is not closed, this point is identical and is a close subpath, instead of skipping this point), but the main thing is in GdipAddPathString – it is necessary to check the type of the _next_ point, and it is necessary to use the correct length to do this. I don't know how easy this would be to test, but it definitely shows up when stroking a text path. Before the fix there are missing segments because the sub-paths for each character don't get closed.

PreferLinux:
OK, I've added a test that covers the first two (in one test, as it is simply checking the bounds of a path created with GdipAddPathString as compared to the result of GdipMeasureString, with suitable tolerances). Unfortunately it doesn't work on the Cairo text rendering backend, because that doesn't support alignment in GdipAddPathString.

hughbe:
Awesome! Thanks - Looks great to me then

filipnavara:
@monojenkins build

akoeplinger:
@monojenkins build

akoeplinger:
@monojenkins whitelist

PR #384, opened 2018-06-18
Pango fixes

PreferLinux:
Some updates to the Pango text backend. The commit messages should be adequate explanation, but if they're not just ask.

### DIFF
--- a/src/font.c
+++ b/src/font.c
@@ -54,6 +54,7 @@ static int ref_familyMonospace = 0;
 static void
 gdip_fontfamily_init (GpFontFamily *fontFamily)
 {
+	fontFamily->collection = NULL;
 	fontFamily->height = -1;
 	fontFamily->linespacing = -1;
 	fontFamily->celldescent = -1;
@@ -374,6 +375,7 @@ GdipGetFontCollectionFamilyList (GpFontCollection *font_collection, INT num_soug
 			return OutOfMemory;
 		}
 
+		gpfamilies[i]->collection = font_collection;
 		gpfamilies[i]->pattern = font_collection->fontset->fonts[i];
 		gpfamilies[i]->allocated = FALSE;
 	}
--- a/src/font.c
+++ b/src/font.c
@@ -780,7 +780,10 @@ gdip_get_pango_font_description (GpFont *font)
 	if (!font->pango) {
 		font->pango = pango_font_description_new ();
 		pango_font_description_set_family (font->pango, (char *)font->face);
-		pango_font_description_set_size (font->pango, font->emSize * PANGO_SCALE);
+		
+		float sizeInPoints = gdip_unit_conversion (font->unit, UnitPoint, gdip_get_display_dpi(), gtMemoryBitmap, font->emSize);
+		
+		pango_font_description_set_size (font->pango, sizeInPoints * PANGO_SCALE);
 
 		if (font->style & FontStyleBold)
 			pango_font_description_set_weight (font->pango, PANGO_WEIGHT_BOLD);
--- a/src/text-pango.c
+++ b/src/text-pango.c
@@ -165,6 +165,7 @@ gdip_process_string (gchar *text, int length, int removeAccelerators, int trimSp
 		iter2 = g_utf8_next_char (iter);
 		g_string_append_len (res, iter, iter2 - iter);
 		j += iter2 - iter;
+		r--;
 	}
 	if (charsRemoved && *charsRemoved && j > 0) {
 		int prevj = (g_utf8_prev_char (res->str + j) - res->str);
--- a/src/text-pango.c
+++ b/src/text-pango.c
@@ -445,7 +445,7 @@ gdip_pango_setup_layout (cairo_t *cr, GDIPCONST WCHAR *stringUnicode, int length
 			pango_layout_iter_get_line_yrange (iter, &y0, &y1);
 			//g_warning("yrange: %d  %d  clipy2: %f", y0 / PANGO_SCALE, y1 / PANGO_SCALE, clipy2);
 			/* StringFormatFlagsLineLimit */
-			if (((fmt->formatFlags & StringFormatFlagsLineLimit) && y1 / PANGO_SCALE >= clipy2) || (y0 / PANGO_SCALE >= clipy2)) {
+			if (((fmt->formatFlags & StringFormatFlagsLineLimit) && y1 / PANGO_SCALE > clipy2) || (y0 / PANGO_SCALE >= clipy2)) {
 				PangoLayoutLine *line = pango_layout_iter_get_line_readonly (iter);
 				pango_layout_set_text (layout, pango_layout_get_text (layout), line->start_index);
 				
--- a/src/text-pango.c
+++ b/src/text-pango.c
@@ -604,7 +604,7 @@ pango_MeasureString (GpGraphics *graphics, GDIPCONST WCHAR *stringUnicode, int l
 			if (iter == NULL)
 				break;
 			pango_layout_iter_get_line_yrange (iter, &y0, &y1);
-			if (y0 / PANGO_SCALE >= max_y)
+			if ((format && (format->formatFlags & StringFormatFlagsLineLimit) && y1 / PANGO_SCALE > max_y) || y0 / PANGO_SCALE >= max_y)
 				break;
 			lines++;
 			if (pango_layout_iter_at_last_line (iter)) {
--- a/src/text-pango.c
+++ b/src/text-pango.c
@@ -596,6 +596,7 @@ pango_MeasureString (GpGraphics *graphics, GDIPCONST WCHAR *stringUnicode, int l
 		int lastIndex;
 		int y0;
 		int y1;
+		int len;
 		double min_x;
 		double max_x;
 		double max_y;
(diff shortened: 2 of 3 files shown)


filipnavara:
@monojenkins build

src/font.c
```diff
@@ -743,7 +745,10 @@ gdip_get_pango_font_description (GpFont *font)
 	if (!font->pango) {
 		font->pango = pango_font_description_new ();
 		pango_font_description_set_family (font->pango, (char *)font->face);
-		pango_font_description_set_size (font->pango, font->emSize * PANGO_SCALE);
+		
+		float sizeInPoints = gdip_unit_conversion (font->unit, UnitPoint, gdip_get_display_dpi(), gtMemoryBitmap, font->emSize);
+		
+		pango_font_description_set_size (font->pango, sizeInPoints * PANGO_SCALE);
```
filipnavara: Unfortunately this won't work on the Pango CoreText backend. There's a bug in Pango where the point size is treated as CoreText point size, which is actually not the typographic point size, but something that HTML calls "px" (device independent pixel). One of them 1/72 inch, the other one is 1/96 inch.

Then again, it was broken before, the bug is in Cairo and not here, and the FreeType backend is currently enforced on all platforms anyway... so just a random note :-)

src/text-pango.c
```diff
@@ -156,9 +156,11 @@ gdip_process_string (gchar *text, int length, int removeAccelerators, int trimSp
...
 	}
 	/* always ensure that at least one space is measured */
-	if (!nonws && trimSpace) {
-		g_string_append_c (res, ' ');
-		j++;
+	if (!nonws && trimSpace && length > 0) {
+		iter = text;
+		iter2 = g_utf8_next_char (iter);
+		g_string_append_len (res, iter, iter2 - iter);
+		j += iter2 - iter;
```
filipnavara: Would it be possible to make simple unit test for this? Just so we can compare the behavior on Windows and Linux.

src/text-pango.c
```diff
@@ -451,7 +453,7 @@ gdip_pango_setup_layout (GpGraphics *graphics, GDIPCONST WCHAR *stringUnicode, i
 			pango_layout_iter_get_line_yrange (iter, &y0, &y1);
 			//g_warning("yrange: %d  %d  clipy2: %f", y0 / PANGO_SCALE, y1 / PANGO_SCALE, clipy2);
 			/* StringFormatFlagsLineLimit */
-			if (((fmt->formatFlags & StringFormatFlagsLineLimit) && y1 / PANGO_SCALE >= clipy2) || (y0 / PANGO_SCALE >= clipy2)) {
+			if (((fmt->formatFlags & StringFormatFlagsLineLimit) && y1 / PANGO_SCALE > FrameHeight) || (y0 / PANGO_SCALE >= clipy2)) {
```
filipnavara: This won't work properly when using the DrawString overloads that take point-only coordinates. After all that's why `clipy2` is there in the first place.

src/font.c
PreferLinux: Would pango_font_description_set_absolute_size() (meant to be the size in device units) solve that? I was in two minds about whether it was better to calculate the point size (which I did in the end), or to use that and the pixel size which is already computed on creating the font...

src/text-pango.c
PreferLinux: Should be, measure a string with, for example, an em-space (the issue I had). I'll have a go tomorrow.

src/font.c
filipnavara: Incidentally it would because the CoreText backend doesn't differentiate them properly, so it would cancel out the bug. I am not sure what is a better option, but I am fine with either one.

src/text-pango.c
PreferLinux: This is only hit when `FrameHeight` is greater than 0 (line 437 / 439), so it shouldn't be a problem. `clipy2` only differs from `FrameHeight` if NoClip is not specified, and in that case is set based on the clip region. However, that specific change relates only to LineLimit, and LineLimit should not clip on the clip region, only on the bounds. Probably something else I should write a unit test for.

filipnavara: Eh, sorry, I missed that. It was me who originally changed it from > to >=. There are some unit tests to cover the general line measurements (https://github.com/mono/libgdiplus/commit/0a27652a0ed9e87d3153b6e03e784d9868a12e1e). If I missed something then a unit test would certainly be welcome. I'm generally trying to get a good unit test coverage for the Pango code to have a more convincing case for making it the default (#269).

filipnavara: I guess my intention was that the pixels at FrameHeight are already not part of the rectangle (the general formula that FrameX + FrameHeight - 1 is the last included pixel) and thus y1 >= FrameHeight is already out of the rectangle.

PreferLinux: True, however from looking at the Pango source, [`pango_layout_iter_get_line_yrange`](https://github.com/GNOME/pango/blob/cdb88930da73bdbeb0e6167825121435d7e5e99f/pango/pango-layout.c#L6393) is effectively [`pango_layout_iter_get_line_extents`](https://github.com/GNOME/pango/blob/cdb88930da73bdbeb0e6167825121435d7e5e99f/pango/pango-layout.c#L6352)`(iter, NULL, logical_rect); ... *y1 = logical_rect->y + logical_rect->height;` Therefore `y1` is actually not included in the line either, and making > correct. I realise the Pango documentation does not make this clear.

src/font.c
PreferLinux: I think I settled on the way I did because of the possibility that the output dpi was different to the display dpi (e.g. for a printer). I realise this does not solve that completely, but it does mean the only unit affected is Pixel. I guess the correct thing to do would be to pass in the dpi to use. That'd require a change in #383 (or moving those changes) to set the resolution on the Graphics object used when drawing to a GraphicsPath.

src/text-pango.c
PreferLinux: Well, I hope I've sorted out counting the characters.

But I did a few checks on Windows (via .Net) as to the behaviour there, and it is weird. To me, it looks like unless there is a line ending, the only whitespace removed from the end are actual spaces. If there is a line ending, all trailing whitespace (considering tab as non-whitespace!) is removed. Furthermore, there is no leaving one space / character behind (gives a width of 0).

src/text-pango.c
PreferLinux: I've just added a test-case for LineLimit (709a5c1).

src/text-pango.c
PreferLinux: I've just added a test case for this (aded2ce). I have also added fixes for the reported number of characters fitted (in two parts, one issue was counting the characters removed (109fe41), the other was in counting the characters in the layout (5b6db10)).

src/text-pango.c
filipnavara: LGTM

filipnavara: Any reason why not keep the comparison to clipy2 (instead of FrameHeight)? If it was intentional then we could drop the whole `clipy2` code, but that would certainly need a test.

PreferLinux: Because if NoClip is not set, clipy2 is based on the clip region. The clip region does not affect LineLimit on Windows – it is possible to have LineLimit set, and a clip region that shows a partial line, as long as the bounding box is large enough to fit the whole line.

clipy2 is used later in the line to eliminate drawing lines that are not visible, irrespective of LineLimit.

filipnavara: That still sounds wrong. Imagine this example:
- The clip region would cut of last two lines
- The bounding box has space for both of the lines
You say that StringFormatFlagsLineLimit doesn't take the clipping region into account, but this loop would still break on the second condition and report wrong results.

PreferLinux: Since I was trying to get LineLimit working correctly I never really thought about the second condition except to figure out roughly what it was meant to do. So I never thought about the implications of it (i.e. altering bounds, number of lines / characters fitted). But you're absolutely right. That too should use FrameHeight, and we could indeed get rid of the clipy2 code.

src/text-pango.c
```diff
@@ -435,23 +438,14 @@ gdip_pango_setup_layout (GpGraphics *graphics, GDIPCONST WCHAR *stringUnicode, i
...
 		iter = pango_layout_get_iter (layout);
 		do {
 			if (iter == NULL)
 				break;
 			pango_layout_iter_get_line_yrange (iter, &y0, &y1);
-			//g_warning("yrange: %d  %d  clipy2: %f", y0 / PANGO_SCALE, y1 / PANGO_SCALE, clipy2);
+			//g_warning("yrange: %d  %d  FrameHeight: %f", y0 / PANGO_SCALE, y1 / PANGO_SCALE, FrameHeight);
 			/* StringFormatFlagsLineLimit */
-			if (((fmt->formatFlags & StringFormatFlagsLineLimit) && y1 / PANGO_SCALE >= clipy2) || (y0 / PANGO_SCALE >= clipy2)) {
+			if (((fmt->formatFlags & StringFormatFlagsLineLimit) && y1 / PANGO_SCALE > FrameHeight) || (y0 / PANGO_SCALE >= FrameHeight)) {
```
PreferLinux: OK, changed the second contition to use FrameHeight too, and eliminated clipy2 and co.

Looking further, clipy2 and co were of very limited reliability for MeasureString because when used from managed code, the layout rectangle will have either the location or the size, not both. This code only does anything when the layout rectangle has size information, therefore there is a good chance that basing anything on the clip rectangle is meaningless. Drawing would probably not be affected.

filipnavara:
Can you rebase please to resolve the conflict?

PreferLinux:
Done.

PreferLinux:
I'm working on the crash on Mac (involved in counting the number of characters fitted). At the same time, I've spotted a related minor error that I'm fixing as well since I think it makes the fix for the crash cleaner.

PreferLinux:
OK, looks like that is sorted now.

filipnavara:
@monojenkins build

PR #385, opened 2018-06-21
Region fixes

PreferLinux:
Region Union was sometimes doing strange things. The problem seems to be that gdip_intersects was returning true in some situations for rectangles that do not necessarily intersect (right of first == X of next, irrespective of Y / Height). I have added a test-case for this.

I initially made gdip_intersects return true for any rectangles that touch, as this would seem to be the desired behaviour for unions. However, that caused test failures for XOR which traces back to exclude, so I tried changing back to only returning true when actually overlapping. That in turn caused test failures for union, so I added a parameter to choose which behaviour was desired.

### DIFF
--- a/src/region.c
+++ b/src/region.c
@@ -247,18 +247,21 @@ gdip_is_InfiniteRegion (const GpRegion *region)
 }
 
 static BOOL
-gdip_intersects (const GpRectF *rect1, const GpRectF *rect2, BOOL includeTouching)
+gdip_intersects (const GpRectF *rect1, const GpRectF *rect2)
 {
-	if (includeTouching)
-		return (rect1->X <= rect2->X + rect2->Width &&
-			rect1->X + rect1->Width >= rect2->X &&
-			rect1->Y <= rect2->Y + rect2->Height &&
-			rect1->Y + rect1->Height >= rect2->Y);
-
 	return (rect1->X < rect2->X + rect2->Width &&
 		rect1->X + rect1->Width > rect2->X &&
 		rect1->Y < rect2->Y + rect2->Height &&
 		rect1->Y + rect1->Height > rect2->Y);
+}
+
+static BOOL
+gdip_intersects_or_touches (GpRectF *rect1, GpRectF *rect2)
+{
+	return (rect1->X <= rect2->X + rect2->Width &&
+		rect1->X + rect1->Width >= rect2->X &&
+		rect1->Y <= rect2->Y + rect2->Height &&
+		rect1->Y + rect1->Height >= rect2->Y);
 }
 
 /* Is source contained in target ? */
@@ -782,7 +785,7 @@ gdip_combine_exclude (GpRegion *region, GpRectF *rtrg, int cntt)
 		/* Current rect with lowest y and X against the target ones */
 		for (i = 0, recttrg = alltrgrects; i < alltrgcnt; i++, recttrg++) {
 
-			if (gdip_intersects (&current, recttrg, FALSE) == FALSE
+			if (gdip_intersects (&current, recttrg) == FALSE
 				|| gdip_equals (&current, recttrg) == TRUE ||
 				recttrg->Height < 0 || recttrg->Width < 0) {
 				continue;
@@ -991,7 +994,7 @@ gdip_combine_union (GpRegion *region, GpRectF *rtrg, int cnttrg)
 				continue;
 			}
 
-			if (gdip_intersects (&current, recttrg, TRUE) == FALSE
+			if (gdip_intersects_or_touches (&current, recttrg) == FALSE
 				|| gdip_equals (&current, recttrg) == TRUE ||
 				recttrg->Height < 0 || recttrg->Width < 0) {
 				continue;
--- a/src/region.c
+++ b/src/region.c
@@ -470,22 +470,20 @@ gdip_getlowestrect (GpRectF *rects, int cnt, GpRectF* src, GpRectF* rslt)
 	return TRUE;
 }
 
-/* Finds a rect that has the lowest x and y after the src rect provided */
+/* Finds the non-empty rect that has the lowest x and y after the current index */
 static BOOL
-gdip_getlowestrect_sorted (GpRectF *rects, int cnt, int* index, GpRectF* src, GpRectF* rslt)
+gdip_getlowestrect_sorted (GpRectF *rects, int cnt, int* index, GpRectF* rslt)
 {
 	GpRectF *current;
 	BOOL found = FALSE;
 
+	(*index)++;
 	for (current = rects + *index; *index < cnt; (*index)++, current++) {
 		if (current->Width <= 0 || current->Height <= 0)
 			continue;
 
-		if (current->Y > src->Y ||
-			(current->Y == src->Y && current->X > src->X)) {
-			found = TRUE;
-			break;
-		}
+		found = TRUE;
+		break;
 	}
 
 	if (found == FALSE) {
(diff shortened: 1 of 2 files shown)


filipnavara:
This doesn't sound right and I believe it may only mask the real issue. Rectangles are quite well defined not to include the bottom right corner (ie. the last included pixel is X + Width - 1, Y + Height - 1 for Width > 0 and Height > 0). This makes the original definition of `gdip_intersects` correct. If the code in the `union` operation after the check misbehaves then it should definitely be fixed, but this looks like you are only touching an edge case.

PreferLinux:
You're probably right, but the real issue was the `if (rect1->X + rect1->Width == rect2->X) { return TRUE; }` in `gdip_intersects`. That simply can't be correct, because they can be separate vertically.

filipnavara:
Yeah, that one is definitely wrong :-)

PreferLinux:
The issue with `union` will be that it is even checking for an intersection – it needs to check for rectangles that can be merged, and that includes ones which do not actually intersect but only touch. So yes, that part of the issue is not with `gdip_intersects` but with `union`. Probably the best way to deal with this is to split my method into two, keeping `gdip_intersects` as you say it should be, and having a new method checking for touching rectangles. It may be more complicated than that too, because `union` still doesn't join some rectangles that it should.

filipnavara:
Can you submit a PR for the gdip_intersects change that removes the obviously wrong line, please?

PreferLinux:
I can do, but doing only that would cause two Union test-case failures (Rect + No Intersect Left on line 1494 and Rect + No Intersect Right on line 1508).

filipnavara:
I guess it still makes more sense to temporarily disable the tests and fix the obvious bug. Second opinion, @hughbe?

filipnavara:
@monojenkins build

filipnavara:
Yeah, there's a LOT of inefficient code and this is one of the worse parts.

Today, `libgdiplus` is basically maintained by volunteers in their spare time. Up to about year ago the test suite was basically non-existent and the only reasonable test coverage was from `System.Drawing` tests in Mono. The downside of those tests is that they are quite far away from the tested code and the test - fix - test again cycle is taking too long. Additionally, it was quite hard to measure some code coverage. Doing any big changes was very risky and it took months to get someone to review them and merge. Things are changing for the better though. @hughbe did a HUGE work on building the unit tests, which made it possible to replace large chunks of code with a reasonable peace of mind.

filipnavara:
Perfect, thanks!

PreferLinux:
There are already tests there for rectangles touching at the corners, just above the test I added. So all good on that front.

Also, thank you to both you and @hughbe for the great work you're doing here!

PreferLinux:
Hum, I only noticed the Travis CI failure on the Mono System.Drawing tests last night... With three intersecting rectangles, when doing the second union it is finding the final part of the second rectangle (which touches) before finding the new rectangle that's above it. The immediate problem is that union assumes the current region scan rectangles will be rejected by intersects when they no longer are (they touch), but the real problem is the middle loop assuming the rectangles are at least partially sorted when there is actually no guarantee of that.

The fix I've come up with is to sort the array of rectangles (`allrects`). This also allows some optimisations. (Makes the outer loop itself O(n) instead of O(n^2), removes an inner loop, and makes the middle loop itself and remaining inner loop approach O(1). The middle loop still has some O(n) operations.) I've created new helper methods where appropriate (`gdip_getlowestrect_sorted`, a comparison method, two sorts, a binary search, `gdip_add_rect_to_array_sorted`), and use the current location in the array where possible.

I have copied some of those tests that were failing into here too.

And in examining that, I've found another union bug that crops up when rectangles are arranged like this:
```
 ________
|   __   |
|__|  |__|
   |  |
   |__|
```
The top rectangle gets the height of the taller one, which in this case is wrong (line 924). I've got a fix for that (and added a test for it), and also for #340 (and removed the `#if` that blocked the tests for that).

I've also changed `gdip_add_rect_to_array` to take a capacity. This does two things: allows specifying the initial number of elements (so that doesn't need to resize when populating it initially, changing from O(n^2) to O(n)), and allows increasing the size significantly (instead of just +1) if the array doesn't have enough capacity. When storing the rectangles in the Region I first copy them to an exact-length array (using a helper method).

I need to tidy up the tests added and anything else I spot before I commit and push the changes. Any comments welcome, especially as I realise that at least the last of those changes is fairly major.

filipnavara:
Wow, I was about to do the sorting myself, but didn't get time for it yet :-) I am glad you took up the challenge!

As far as reallocation goes there's a way to predict the worst case scenario and allocate big enough array of rectangles upfront. It could be shrunk back at the end of the function if necessary. This should make all the out-of-memory errors easier to handle and reduce reallocations in most cases.

PreferLinux:
Well, I've only done it for union so far, but it should be easy enough to extend it to the other combines (especially as exclude is the only other big one). There may be more easy optimisations possible yet too.

Good point. So far I've just allocated the initial length before initially adding contents and let the expanding handle anything added later, but that's really easy to change.

PreferLinux:
Rebased to fix merge conflict.

filipnavara:
Looks like it passed the CI this time (the Windows error is unrelated). I will review it soon.

src/region.c
```diff
@@ -337,6 +461,33 @@ gdip_getlowestrect (GpRectF *rects, int cnt, GpRectF* src, GpRectF* rslt)
...
+
+	for (current = rects + *index; *index < cnt; (*index)++, current++) {
+		if (current->Width <= 0 || current->Height <= 0)
+			continue;
+
+		if (current->Y > src->Y ||
+			(current->Y == src->Y && current->X > src->X)) {
+			found = TRUE;
+			break;
+		}
```
filipnavara: Won't this always be true?

Or, put another way, how are two rectangles with the same origin handled? Eg. merging {4, 4; 100x100} with {4, 4; 200x200}.

src/region.c
```diff
@@ -53,25 +53,144 @@ gdip_region_new ()
...
+	}
+	next = array + lower;
+	if (index)
+		*index = lower;
+	return next;
+}
+
+static GpStatus
+gdip_add_rect_to_array_sorted (GpRectF** srcarray, int* elements, int* capacity, const GpRectF* rect)
+{
```
filipnavara: This function is not used as far as I can tell. It would certainly make sense to use it at the place where `gdip_sort_rect_array_sorted` is used now...

PreferLinux: Correct. I was using it there initially, but changed to using the sort and didn't remove the method. In union the modifications to `recttrg` require sorting after them, so I decided it would probably be better to do one sort that can handle both of those. However, on thinking about it again, only one of the two ever happens (either `rslt` is empty, or the remnant of `recttrg` will be empty). Therefore it would be better to use this there as you suggest.

src/region.c
PreferLinux: Yes, assuming nothing sets a rectangle's location to (0, 0) (something that union used to do, which I've removed) and then sorts the array.

As long as there are only two rectangles with the same origin it is fine, and it shouldn't be possible to have more than two (would require a region made from specially-crafted region data). This'll get the first, and then the middle loop will get the second one. But still something worthy of thought.

filipnavara: Right, so the same origin case is handled properly. Then it looks like this check should not be necessary anymore.

filipnavara: You can actually get rid of `gdip_getlowestrect_sorted` altogether and replace it with simple loop like this: https://github.com/filipnavara/libgdiplus/commit/e3296d45ac07d6e4a1a68444430b07fb21da6ec0

PreferLinux: That looks excellent. How do we include that?

PreferLinux:
Included @filipnavara's change, and also rebased to fix the merge conflict.

PreferLinux:
Rebased to fix merge conflict, and updated one test because of test rectangles being renamed from `noIntersect___` to `touching___`

PR #386, opened 2018-06-26
Initial AIX support for libgdiplus

NattyNarwhal:
Tested with a mix of AIX Toolbox and Perzl packages. Builds a 64
bit libgdiplus to go with mono.

X11, giflib (but not ungiflib), libexif, libpng, giflib, and libtiff
are all used.

Due to deficiencies in the AIX Toolbox version of Cairo, where it
specifies system X11 in the package, but IBM's system X11 lacks
a package, you must specify Cairo flags manually to the configure
script like so. This isn't done by the script as you may be using
a Cairo linked against a newer X.org instead.

```shell
CAIRO_CFLAGS=-I/opt/freeware/include/cairo
CAIRO_LIBS="-L/opt/freeware/lib -lcairo"
CPPFLAGS=-I/opt/freeware/include/cairo
```

This is untested on PASE.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -119,6 +119,17 @@ case "$host" in
 		CFLAGS="-pthreads $CFLAGS"
 		;;
 
+	*-*-aix*|*-*-os400)
+		AC_DEFINE_UNQUOTED(AIX,1,[AIX or PASE])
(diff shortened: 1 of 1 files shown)


akoeplinger:
@monojenkins build

Issue #401, opened 2018-08-04
GdipCreatePath2 should validate the type array


hughbe:
There's a FIXME in the source code
```
	/* FIXME: Check whether path types are valid before adding them. MS does
	 * some checking and does not use the points passed in, if they look
	 * invalid. Some cases are like last type being PathPointTypeStart or
	 * PathPointTypeCloseSubpath etc.
	 */

```

Issue #400, opened 2018-08-04
GdipCreatePath2 should normalize the type array


hughbe:
All start types are converted to 0 - i.e. the first element of the type array is set to zero no matter what is passed

PR #410, opened 2018-08-05
Fix GdipCreatePath2 validation and add tests

hughbe:
Fixes https://github.com/mono/libgdiplus/issues/401
Fixes https://github.com/mono/libgdiplus/issues/400

### DIFF
--- a/src/general.c
+++ b/src/general.c
@@ -227,15 +227,15 @@ gdip_get_display_dpi ()
 GpPointF *
 convert_points (const GpPoint *point, int count)
 {
-	int i;
+	if (!point || count < 0)
+		return NULL;
+
 	GpPointF *retval = (GpPointF *) GdipAlloc (sizeof (GpPointF) * count);
 	if (!retval)
 		return NULL;
 
-	for (i = 0; i < count; i++) {
-		retval [i].X = (float) point [i].X;
-		retval [i].Y = (float) point [i].Y;
-	}
+	for (int i = 0; i < count; i++)
+		gdip_PointF_from_Point (&point[i], &retval[i]);
 
 	return retval;
 }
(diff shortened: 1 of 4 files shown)


hughbe:
Rebased /cc @marek-safar @filipnavara

hughbe:
Friendly ping @marek-safar @fillipnavara - I’m back with a laptop today and have some PRS that depend on this one I fancy sending in

src/general.c
```diff
@@ -227,28 +227,30 @@ gdip_get_display_dpi ()
 GpPointF *
 convert_points (const GpPoint *point, int count)
 {
-	int i;
+	if (!point || count < 0)
```
filipnavara: I'd suggest `count <= 0` since it would return NULL anyway in the next step.

src/general.c
```diff
@@ -227,28 +227,30 @@ gdip_get_display_dpi ()
...
 
 	return retval;
 }
 
 GpRectF *
 convert_rects (const GpRect *rect, int count)
 {
-	int i;
+	if (!rect || count < 0)
+		return NULL;
```
filipnavara: Likewise, `count <= 0`.

src/graphics-path.c
```diff
@@ -199,6 +199,63 @@ append_curve (GpPath *path, const GpPointF *points, GpPointF *tangents, int offs
...
+gdip_validate_path_types (const BYTE *types, INT count)
+{
+	const BYTE *currentType = types;
+	INT remaining = count;
+
+	if (count == 1)
+		return TRUE;
+
+	while (TRUE) {
+		remaining -= 1;
```
marek-safar: Just use `--` and `++` everywhere

src/general.c
hughbe: Can't change this - on my PC this causes test failures in `testpathgradientbrush`

Will investigate further in another PR?

src/graphics-path.c
hughbe: Fixed :)

src/general.c
filipnavara: That's weird, but I am fine with postponing it.

PR #394, opened 2018-08-03
Fix infinite path region calculations

hughbe:
Should check for Ok, not !Ok

Fixes https://github.com/mono/libgdiplus/issues/399
Fixes https://github.com/mono/libgdiplus/issues/398
Contributes to https://github.com/mono/libgdiplus/issues/395
Contributes to #338

PR #415, opened 2018-08-05
Add more region combine tests

hughbe:
Depends on https://github.com/mono/libgdiplus/pull/394

Now covers cases of Infinite/Empty Rect/Path region combinations and rects that have NO intersections - i.e. are not even touching!


hughbe:
@filipnavara @marek-safar friendly ping :)

PR #418, opened 2018-08-09
Fix Windows build

filipnavara:
bzip.org is down :-/


filipnavara:
@marek-safar Can you merge pls? The CI is broken without it unfortunately.

PR #419, opened 2018-08-13
Fix drawing bezier curves.

PreferLinux:
The old is giving `cairo_curve_to` the origin point and then two control points, with no destination. It needs the two control points and the destination, with the origin being specified by the earlier call to `cairo_move_to`.

### DIFF
--- a/src/graphics-cairo.c
+++ b/src/graphics-cairo.c
@@ -274,7 +274,7 @@ cairo_DrawBeziers (GpGraphics *graphics, GpPen *pen, GDIPCONST GpPointF *points,
 	/* We use graphics->copy_of_ctm matrix for path creation. We should have it set already. */
 	gdip_cairo_move_to (graphics, points [0].X, points [0].Y, TRUE, TRUE);
 
-	for (i = 0, j = 1, k = 2; i < count - 3; i += 3, j += 3, k += 3) {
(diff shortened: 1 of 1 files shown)


src/graphics-cairo.c
```diff
@@ -274,7 +274,7 @@ cairo_DrawBeziers (GpGraphics *graphics, GpPen *pen, GDIPCONST GpPointF *points,
 	/* We use graphics->copy_of_ctm matrix for path creation. We should have it set already. */
 	gdip_cairo_move_to (graphics, points [0].X, points [0].Y, TRUE, TRUE);
 
-	for (i = 0, j = 1, k = 2; i < count - 3; i += 3, j += 3, k += 3) {
+	for (i = 1, j = 2, k = 3; i < count - 2; i += 3, j += 3, k += 3) {
```
filipnavara: I would have probably preferred `k < count` instead of `i < count - 2` because it's easier read. Otherwise looks good to me.

PreferLinux: I agree. I just went with what was there and didn't really think about that side of things too much.

PreferLinux:
Changed to checking `k < count` instead of `i < count - 2` as suggested.

hughbe:
Do you think this fixes https://github.com/mono/libgdiplus/issues/423?

PR #421, opened 2018-08-28
Transform fixes

PreferLinux:
Fix setting the world transform when in a container. It needs to take the parent transform into account.

Also store and use the original GpGraphics bounds, to fix the action of `apply_world_to_bounds` (was applying the complete transform to partially-transformed coordinates.


migueldeicaza:
I will merge this, and if there are issues with metafiles, we can address those later.   It is a net improvement.

PR #434, opened 2018-08-30
 Fix combine for negative rects/paths and IsEmpty

hughbe:
Fixes https://github.com/mono/libgdiplus/issues/336


migueldeicaza:
This will need to be rebased

PR #442, opened 2018-09-09
Move some region code to common

hughbe:
No functional change, but some PRs will be using this around in graphics/graphics-path etc so its good to move to common


migueldeicaza:
This will need to be rebased

PR #444, opened 2018-09-10
 Rename BitmapData->ActiveBitmapData

hughbe:
To represent the difference between GDI+ and libgdiplus

This also lays the groundwork for fixing the definition of BitmapData to match GDI.


migueldeicaza:

PR #474, opened 2018-09-13
 Add more Pen tests, fix bugs and cleanup the source

hughbe:
Fixes https://github.com/mono/libgdiplus/issues/473

### DIFF
--- a/src/pen.c
+++ b/src/pen.c
@@ -252,12 +252,12 @@ GdipCreatePen1 (ARGB argb, REAL width, GpUnit unit, GpPen **pen)
 	if (!gdiplusInitialized)
 		return GdiplusNotInitialized;
 
-	if (!pen || unit < UnitWorld || unit > UnitCairoPoint || unit == UnitDisplay)
+	if (!pen || unit > UnitCairoPoint || unit == UnitDisplay)
 		return InvalidParameter;
 
 	result = gdip_pen_new ();
 	if (!result)
-		goto error;
+		return OutOfMemory;
 
 	result->color = argb;
 	result->width = width;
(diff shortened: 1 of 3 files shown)


src/pen.c
```diff
@@ -252,31 +252,26 @@ GdipCreatePen1 (ARGB argb, REAL width, GpUnit unit, GpPen **pen)
...
 		return GdiplusNotInitialized;
 
-	if (!pen || unit < UnitWorld || unit > UnitCairoPoint || unit == UnitDisplay)
+	if (!pen || unit > UnitCairoPoint || unit == UnitDisplay)
 		return InvalidParameter;
 
 	result = gdip_pen_new ();
 	if (!result)
-		goto error;
+		return OutOfMemory;
```
filipnavara: This leaves *pen uninitialized, which is not a good practice.

src/pen.c
```diff
@@ -252,31 +252,26 @@ GdipCreatePen1 (ARGB argb, REAL width, GpUnit unit, GpPen **pen)
...
 	result->color = argb;
 	result->width = width;
 	result->unit = unit;
 	result->own_brush = TRUE;
 
 	status = GdipCreateSolidFill (argb, (GpSolidFill **) &result->brush);
-	if (status != Ok)
-		goto error;
+	if (status != Ok) {
+		GdipDeletePen (result);
```
filipnavara: Same here.

filipnavara: ...and in GdipCreatePen2 below as well.

hughbe:
Well actually I think Windows would crash in these situations (but we can't actually reach it realistically because these allocations wouldn't really fail normally). So we're free to do whatever we want, so I added these assignments back

PR #475, opened 2018-09-13
Add more TextureBrush tests, fix bugs and cleanup the source

hughbe:


migueldeicaza:
I am approving many of these changes, for the sake of expedience, but I noticed that there are lots of changes from float -> REAL and same for INT.

Is there a rationale for that?

hughbe:
To match the windows definitions in GDI+. I felt it is also a bit cleaner because it makes the intent a bit more clear too. I wouldn’t submit those changes on their own tho

Issue #476, opened 2018-09-13
GdipStringFormatGetGenericTypographic/GdipStringFormatGetGenericDefault should return singletons


hughbe:
Created on demand, deleted when shutting down GDI+

PR #477, opened 2018-09-13
Make generic string formats singletons

hughbe:
Fixes https://github.com/mono/libgdiplus/issues/476

### DIFF
--- a/src/stringformat.c
+++ b/src/stringformat.c
@@ -27,15 +27,26 @@
 #include "stringformat-private.h"
 #include "general-private.h"
 
-/* Generic string formats */
-#if GLIB_CHECK_VERSION(2,32,0)
-static GMutex generic;
-#else
-static GStaticMutex generic = G_STATIC_MUTEX_INIT;
-#endif
-
 static GpStringFormat *stringFormatDefault = NULL;
 static GpStringFormat *stringFormatTypographic = NULL;
+
+GpStatus
+gdip_create_generic_stringformats ()
+{
+	GpStatus status;
+	const int typographicFlags = StringFormatFlagsNoFitBlackBox | StringFormatFlagsLineLimit | StringFormatFlagsNoClip;
+
+	status = GdipCreateStringFormat (0, 0, &stringFormatDefault);
+	if (status != Ok)
+		return status;
+	
+	status = GdipCreateStringFormat (typographicFlags, 0, &stringFormatTypographic);
+	if (status != Ok)
+		return status;
+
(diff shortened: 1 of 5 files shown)


src/stringformat.c
```diff
@@ -82,18 +106,71 @@ GdipCreateStringFormat (INT formatAttributes, LANGID language, GpStringFormat **
...
+		return InvalidParameter;
+
+	if (!gdiplusInitialized)
+		return GdiplusNotInitialized;
+
+#if GLIB_CHECK_VERSION(2,32,0)
+	g_mutex_lock (&generic);
+#else
+	g_static_mutex_lock (&generic);
+#endif
```
filipnavara: Not a big fan of the mutexes. I wonder if it was easier to just allocate it during the library initialization and then simply return it here.

hughbe: Sure - this will have a perf/memory impact though but its tiny (like 20 bytes or something)

src/stringformat.c
```diff
@@ -27,6 +27,41 @@
...
+	if (status != Ok)
+		return status;
+	
+	status = GdipCreateStringFormat (typographicFlags, 0, &stringFormatTypographic);
+	if (status != Ok)
+		return status;
+
+	stringFormatTypographic->trimming = StringTrimmingNone;
+	return Ok;
+}
```
filipnavara: On a second thought, can we make `stringFormatGeneric` and `stringFormatTypographic` declared as `GpStringFormat` instead of `GpStringFormat *`? Then you would not need to allocate and free memory and you can avoid a lot of error checking.

hughbe:
Build continuous-integration/travis-ci/pr

src/stringformat.c
```diff
@@ -44,6 +61,23 @@ gdip_string_format_init (GpStringFormat *result)
...
+	gdip_string_format_init (&stringFormatTypographic);
+	stringFormatTypographic.formatFlags = StringFormatFlagsNoFitBlackBox | StringFormatFlagsLineLimit | StringFormatFlagsNoClip;
+	stringFormatTypographic.trimming = StringTrimmingNone;
+}
+
+void
+gdip_delete_generic_stringformats ()
+{
+	gdip_string_format_clear (&stringFormatDefault);
+	gdip_string_format_clear (&stringFormatTypographic);
```
filipnavara: I don't think this cleanup is necessary since the allocated fields are never set. Or maybe I am incorrectly assuming that these objects are immutable? Not sure if there's a test for that.

PR #478, opened 2018-09-14
Add more ImageAttributes tests and fix validation compat bug

hughbe:


src/imageattributes.c
```diff
@@ -90,10 +90,10 @@ gdip_get_image_attribute (GpImageAttributes* attr, ColorAdjustType type)
 void
 gdip_process_bitmap_attributes (GpBitmap *bitmap, void **dest, GpImageAttributes* attr, BOOL *allocated)
 {
+	GpStatus;
```
filipnavara: What is this for? Shouldn't it be `GpStatus status`?

PR #481, opened 2018-09-14
Cleanup Graphics container/state definitions

hughbe:
To match GDI+ (and so I can write some tests for these in the future)


migueldeicaza:
This will need to be rebased

PR #482, opened 2018-09-14
Fix incorrect cast in apply_world_to_bounds

hughbe:
Taking the address of an array then converting it to a pointer is incorrect.
- not sure why this code actually works though, maybe because PointF the same size as a pointer?

And remove unecessary checks for status
- there's no way GdipTransformMatrixPoints can fail here and no one even checks the return value anyway

### DIFF
--- a/src/graphics.c
+++ b/src/graphics.c
@@ -580,19 +580,16 @@ GdipSaveGraphics (GpGraphics *graphics, unsigned int *state)
 	return Ok;
 }
 
-static GpStatus
+static void
 apply_world_to_bounds (GpGraphics *graphics)
 {
-	GpStatus status;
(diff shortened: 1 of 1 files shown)


filipnavara:

src/graphics.c
```diff
@@ -577,19 +577,16 @@ GdipSaveGraphics (GpGraphics *graphics, unsigned int *state)
 	return Ok;
 }
 
-static GpStatus
+static void
 apply_world_to_bounds (GpGraphics *graphics)
```
filipnavara: There's still `return Ok;` at the end of the function...

migueldeicaza:
This will need to be rebased

Issue #499, opened 2018-09-18
SolidBrush filling with alpha and CompositingModeSourceCopy is incorrect


hughbe:
```cpp
static void drawOverImage (GpImage *image)
{
    GpStatus status;
    GpBitmap *bitmap;
    GpGraphics *graphics;
    GpSolidFill *brush;

    GdipCreateBitmapFromScan0 (256, 256, 0, PixelFormat32bppARGB, NULL, &bitmap);
    GdipGetImageGraphicsContext ((GpImage *) bitmap, &graphics);

    GdipCreateSolidFill (0xFFFF00FF, &brush);
    GdipFillRectangle (graphics, (GpBrush *) brush, 0, 0, 256, 256);

    GdipDrawImage (graphics, image, 0, 0);

    GdipDeleteGraphics (graphics);
    GdipDisposeImage ((GpImage *) bitmap);
    GdipDeleteBrush ((GpBrush *) brush);
}

static void test_fill ()
{
    GpSolidFill *brush;
    GpBitmap *bitmap;
    GpGraphics *graphics;

    GdipCreateBitmapFromScan0 (256, 256, 0, PixelFormat32bppARGB, NULL, &bitmap);
    GdipGetImageGraphicsContext ((GpImage *) bitmap, &graphics);
    GdipSetCompositingMode (graphics, CompositingModeSourceCopy);
    GdipCreateSolidFill (0xAABBCCDD, &brush);
    
    GdipFillRectangle (graphics, brush, 64, 64, 128, 128);
    drawOverImage ((GpImage *) bitmap);

    GdipDeleteBrush ((GpBrush *) brush);
    GdipDeleteGraphics (graphics);
    GdipDisposeImage ((GpImage *) bitmap);
}
```

The resulting image in `drawOverImage` is incorrect as the alpha is `0xFF` in libgdiplus but is not in GDI+.

### GDI+
![gdiplus](https://user-images.githubusercontent.com/1275900/45709229-697d7000-bb7b-11e8-91e8-fc02b5cd9c69.png)

### Libgdiplus
![libgdiplus](https://user-images.githubusercontent.com/1275900/45709233-6d10f700-bb7b-11e8-94a8-4999909b735a.png)

PR #500, opened 2018-09-19
Fix SolidBrush alpha compositing mode bug

hughbe:
Fixes https://github.com/mono/libgdiplus/issues/499

/cc @filipnavara

### DIFF
--- a/src/solidbrush.c
+++ b/src/solidbrush.c
@@ -64,7 +64,6 @@ GpStatus
 gdip_solidfill_setup (GpGraphics *graphics, GpBrush *brush)
 {
 	GpSolidFill *solid;
-	int A, R, G, B;
 
 	if (!graphics || !brush)
 		return InvalidParameter;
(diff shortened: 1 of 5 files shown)


src/solidbrush.c
```diff
@@ -75,26 +74,26 @@ gdip_solidfill_setup (GpGraphics *graphics, GpBrush *brush)
...
+			solid->R = 0;
+			solid->G = 0;
+			solid->B = 0;
+		}
+		else {
+			solid->A = (double) a / 255.0;
+			solid->R = (double) r / 255.0;
+			solid->G = (double) g / 255.0;
+			solid->B = (double) b / 255.0;
+		}
```
filipnavara: What is the reason for special casing `a == 0`?

The thing that I am worried about is that it may mask an issue with premultiplies vs non-premultiplied surfaces. It is probably correct, but it would be nice to have tests for some less special alpha values, such as 0x80.

hughbe: This came up when writing tests - basically if the alpha value is zero, the pixel is set to `0x00000000`. I've been thinking a little about this and taken a look at GDI+. In fact look like GDI+ premultiplies the color of the solid brush regardless of whether the surface format is premultiplied or not... But I'm not really sure how this would integrate with Cairo, considering that cairo doesn't really support premultiplied formats immediately. Also, I may be wrong but I **think** that GDI+ is all premultiplied deep in its implementation (from taking a look at the decompilation).

screen shot 2018-09-19 at 10 34 03


Anyway: I've special cased these tests for GDI+/libgdiplus but haven't really taken an issue because the colors are so close the difference is not even noticeable. We already have a test for less special alpha numbers: e.g.
```cpp
    // Fill with brush - CompositingModeSourceCopy.
    status = GdipSetSolidFillColor (brush, 0xDDA03040);
    assertEqualInt (status, Ok);

    status = GdipSetCompositingMode (graphics, CompositingModeSourceCopy);
    assertEqualInt (status, Ok);

    status = GdipFillRectangle (graphics, brush, 0, 0, 1, 1);
    assertEqualInt (status, Ok);

    status = GdipBitmapGetPixel (bitmap, 0, 0, &color);
    assertEqualInt (status, Ok);
#if defined(USE_WINDOWS_GDIPLUS)
    assertEqualARGB (color, 0xFF8B2A37);
#else
    assertEqualARGB (color, 0xFF8B2937);
#endif
```

as you can see the colors are literally off by like 1 byte in the blue channel

hughbe: However, that is for surfaces without an alpha channel. With an alpha channel, the difference is greater. E.g.:

```cpp
	// PixelFormat32bppARGB.
    status = GdipCreateBitmapFromScan0 (256, 256, 0, PixelFormat32bppARGB, NULL, &bitmap);
    assertEqualInt (status, Ok);

    status = GdipGetImageGraphicsContext ((GpImage *) bitmap, &graphics);
    assertEqualInt (status, Ok);

    status = GdipBitmapSetPixel (bitmap, 0, 0, 0x00000000);
    assertEqualInt (status, Ok);
	
    status = GdipSetSolidFillColor (brush, 0xDDA03040);
    assertEqualInt (status, Ok);
    // Fill with brush - CompositingModeSourceCopy.
    status = GdipSetSolidFillColor (brush, 0xDDA03040);
    assertEqualInt (status, Ok);

    status = GdipSetCompositingMode (graphics, CompositingModeSourceCopy);
    assertEqualInt (status, Ok);

    status = GdipFillRectangle (graphics, brush, 0, 0, 1, 1);
    assertEqualInt (status, Ok);

    status = GdipBitmapGetPixel (bitmap, 0, 0, &color);
    assertEqualInt (status, Ok);
#if defined(USE_WINDOWS_GDIPLUS)
    assertEqualARGB (color, 0xDDA0303F);
#else
    assertEqualARGB (color, 0xDD8B2937);
#endif
```

filipnavara: > But I'm not really sure how this would integrate with Cairo, considering that cairo doesn't really support premultiplied formats immediately.

Most of the Cairo _surfaces_ are actually premultiplied (at least as used by libgdiplus), which should match whatever GDI+ is doing.

I will have to think about it more though. There are way too many bugs triggered by the premultiplation done incorrectly or at wrong places. I'll add it to my backlog to investigate this further, but unfortunately I am quite swamped these days. :-/

hughbe:
@filipnavara friendly ping :)

filipnavara:
Looks good to me now.

PR #516, opened 2018-11-22
fix a potential double free issue

markusbeth:
Clear flag GBD_OWN_SCAN0 if the memory for scan0 is not allocated within
the function GdipBitmapLockBits. Failing to do so leads to an unintended
free() in GdipBitmapUnlockBits.

Fixes: 515


akoeplinger:
@monojenkins build

PR #517, opened 2018-11-22
fix shell syntax error in configure.ac

markusbeth:


akoeplinger:
@monojenkins build

configure.ac
```diff
@@ -469,7 +469,7 @@ AS_IF(
 AM_CONDITIONAL([HAS_X11], [test x$x11_available = xyes])
 
 # Add xrender here so that we don't fail to find glib if we won't have xrender.pc
-if test "x$no_x" != "xyes" && "x$x11_available" = "xyes"; then
+if test "x$no_x" != "xyes" && test "x$x11_available" = "xyes"; then
    GDIPLUS_LIBS="$GDIPLUS_LIBS `$PKG_CONFIG --libs xrender `"
    GDIPLUS_CFLAGS="$GDIPLUS_CFLAGS `$PKG_CONFIG --cflags xrender `"
 fi
```
akoeplinger: Looks like this was introduced by https://github.com/mono/libgdiplus/commit/e6f327363358eff2bdf5fe54621d1991598df689 but given that nothing broke and this xrender dependency doesn't seem to be needed I'd rather avoid reintroducing it.

Can you please just delete this whole block instead?

markusbeth: I added a commit to clean up the xrender dependency

Issue #541, opened 2019-04-26
Clone bitmap using PixelFormat.Format4bppIndexed results in Format32bppArgb


gfoidl:
Repro:
```c#
var originMap = new Bitmap(1, 1);
var cloneMap = originMap.Clone(new Rectangle(0, 0, 1, 1), PixelFormat.Format4bppIndexed);
Console.WriteLine("PixelFormat of cloneMap: " + cloneMap.PixelFormat.ToString());
```

Results on linux in `PixelFormat of cloneMap: Format32bppArgb`.
Mono and .NET Core 3.0 show the same behavior.

Expected ist `Format4bppIndexed`.
.NET Core on windows shows the expected result.

Reference: https://github.com/dotnet/corefx/issues/37209

gfoidl:
Hm, it seems that the code in _bitmap.c_ is correct, as https://github.com/mono/libgdiplus/blob/dba6fc63e3559106dde5c0eca58bd6d68436f28f/src/bitmap.c#L1198 sets the pixel format.

Simple test passes.
```c
#include <assert.h>

#include "bitmap-private.h"
//-----------------------------------------------------------------------------
static ULONG_PTR s_token;
//-----------------------------------------------------------------------------
void checkStatus(GpStatus status);
//-----------------------------------------------------------------------------
int main()
{
    GpStatus status;

    GdiplusStartupInput startupInput;
    GdiplusStartupOutput startupOutput;
    startupInput.GdiplusVersion = 2;

    status = GdiplusStartup(&s_token, &startupInput, &startupOutput);
    checkStatus(status);

    GpBitmap* bitmap;
    status = GdipCreateBitmapFromScan0(1, 1, 0, PixelFormat4bppIndexed, NULL, &bitmap);
    checkStatus(status);

    GpBitmap* clone;
    status = GdipCloneBitmapAreaI(0, 0, 1, 1, PixelFormat4bppIndexed, bitmap, &clone);
    checkStatus(status);

    PixelFormat actual;
    status = GdipGetImagePixelFormat(clone, &actual);
    checkStatus(status);

    assert(PixelFormat4bppIndexed == actual);

    GdipDisposeImage(bitmap);
    GdipDisposeImage(clone);
    GdiplusShutdown(s_token);
}
//-----------------------------------------------------------------------------
void checkStatus(GpStatus status)
{
    if (status == Ok) return;

    GdiplusShutdown(s_token);
}
```

gfoidl:
[...]

PR #542, opened 2019-04-26
Added test for cloning a bitmap 

gfoidl:
...and checking if pixel format is set correct.

Verification test for https://github.com/mono/libgdiplus/issues/541


gfoidl:
The test is wrong...sorry. I'll close this one (and open a new one when I can come up with a fix).

PR #543, opened 2019-04-26
[WIP] GdipCloneBitmapAreaI sets pixel format correct

gfoidl:
The pixel format was set from the source-bitmap, not from the argument passed to the method.

Fixes https://github.com/mono/libgdiplus/issues/541
Fixes https://github.com/mono/mono/issues/9903

### DIFF
--- a/src/bitmap.c
+++ b/src/bitmap.c
@@ -1118,6 +1118,7 @@ GdipCloneBitmapAreaI (INT x, INT y, INT width, INT height, PixelFormat format,
 	}
 
 	result->image_format = original->image_format;
+	result->active_bitmap->pixel_format = format;
 
 	status = gdip_bitmap_clone_data_rect (original->active_bitmap, &sr, result->active_bitmap, &dr);
 	if (status != Ok) {
@@ -1183,7 +1184,6 @@ gdip_bitmap_clone_data_rect (ActiveBitmapData *srcData, Rect *srcRect, ActiveBit
 	if (destData->scan0 == NULL) {
 		dest_components = gdip_get_pixel_format_components (srcData->pixel_format);
 		dest_depth = gdip_get_pixel_format_depth (srcData->pixel_format);
-		destData->pixel_format = srcData->pixel_format;
 
 		destData->stride = ((destRect->Width * dest_components * dest_depth) >> 3);
 		gdip_align_stride (destData->stride);
@@ -1195,7 +1195,6 @@ gdip_bitmap_clone_data_rect (ActiveBitmapData *srcData, Rect *srcRect, ActiveBit
 		
 		destData->width = destRect->Width;
 		destData->height = destRect->Height;
-		destData->pixel_format = srcData->pixel_format;
 		destData->reserved = GBD_OWN_SCAN0;
 
 		if (srcData->palette) {
--- a/tests/testbitmap.c
+++ b/tests/testbitmap.c
@@ -758,6 +758,25 @@ static void test_createBitmapFromGraphics ()
 	GdipDeleteGraphics (graphicsWithResolution);
 }
 
+static void test_cloneBitmap()
+{
+	GpStatus status;
+	GpBitmap *bitmap;
+	GpBitmap *clone;
+	
+	status = GdipCreateBitmapFromScan0(1, 1, 0, PixelFormat32bppARGB, NULL, &bitmap);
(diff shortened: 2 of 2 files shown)


src/bitmap.c
```diff
@@ -1118,6 +1118,7 @@ GdipCloneBitmapAreaI (INT x, INT y, INT width, INT height, PixelFormat format,
 	}
 
 	result->image_format = original->image_format;
+	result->active_bitmap->pixel_format = format;
```
filipnavara: I am not sure if this is correct fix. The code below takes the `cairo_format` from the original bitmap (`result->cairo_format = original->cairo_format;` and it seems like it can result in inconsistent values.

gfoidl: Hm, that seems possible. So the `result->cairo_format` should be updated too? To have a consistent state.

Or is there any other way to fix this behavior, which I'm not aware atm?

filipnavara: It's more difficult than that I am afraid. At line 1185 the assumption about destination format == source format is still taken into account. I don't see any code doing the actual format conversion.

filipnavara: Besides that, `PixelFormatDontCare` is not handled either. In that case it should use the source format.

tests/testbitmap.c
```diff
@@ -758,6 +758,25 @@ static void test_createBitmapFromGraphics ()
...
+	status = GdipCreateBitmapFromScan0(1, 1, 0, PixelFormat32bppARGB, NULL, &bitmap);
+	assertEqualInt(status, Ok);
+
+	status = GdipCloneBitmapAreaI(0, 0, 1, 1, PixelFormat4bppIndexed, bitmap, &clone);
+	assertEqualInt(status, Ok);
+
+	PixelFormat actual;
+	status = GdipGetImagePixelFormat(clone, &actual);
+	assertEqualInt(status, Ok);
+
```
filipnavara: I assume it would make sense to test the actual contents of the bitmap here. You can create the 32bpp bitmap from some source data like `{ 0xFFFFFFFF, 0, 0, 0xFFFFFFFF }` with size `2, 2` and then verify the pixel colors on the output (already done on some other tests).

gfoidl: Thanks for the hint. Will update the test.

gfoidl:
> It's more difficult than that I am afraid

After some investigation of the code I think so, too -- unfortunately.
Missing points:
* format conversion in the data
* for indexed formats initialization of the palette, so that [GdipBitmapGetPixel et. al.](https://github.com/mono/libgdiplus/blob/dba6fc63e3559106dde5c0eca58bd6d68436f28f/src/bitmap.c#L2064) works
* ...

I'll try to address these points, but my knowledge of this code-base is somewhat limited, so I can't guarantee any success.

Marking this PR as [WIP] until code is fixed and tests passes.

gfoidl:
Just saw https://github.com/mono/libgdiplus/blob/dba6fc63e3559106dde5c0eca58bd6d68436f28f/src/bitmap.c#L1313 although the function isn't used by clone, it's only used by `gdip_bitmap_change_rect_pixel_format` which in turn is used by lock/unlock bits.

Does this mean that pixel format conversion to indexed formats isn't supported by libgdiplus by design?

filipnavara:
Quite possibly yes. It's difficult to implement and I didn't cross-check with Wine/Windows to see whether it is implemented there.

gfoidl:
On windows and / or .NET cloning works as expected, so it seems that it is implemented there.
At least `clone.GetPixel(x, y)` returns the expected color and pixelformat is set to the given argument.

Maybe it's unrelated, but https://github.com/mono/libgdiplus/blob/dba6fc63e3559106dde5c0eca58bd6d68436f28f/src/bitmap.c#L2149 indicates that there's no `cairo_surface_t` for indexed formats?

filipnavara:
The formats of the Cairo surface are not always exact match to the GDI+ bitmaps. There is some support for the indexed formats, eg. at the beginning of `GdipCreateBitmapFromScan0`.

gfoidl:
Well, I'm not really capable for bringing this PR to a proper end, as my knowledge here is limited. If anyone could come up with a better and proper solution I'd be happy.

gfoidl:
I'm closing as of https://github.com/mono/libgdiplus/pull/543#issuecomment-497703233 and that this PR doesn't sit around opened without an end.

Issue #443, opened 2018-09-10
GdipBitmapSetPixel when x/y is the width/height crashes


hughbe:
Out of bounds access to memory

PR #544, opened 2019-04-27
GdipBitmapSetPixel correct bounds-check

gfoidl:
Fixes https://github.com/mono/libgdiplus/issues/443

The condition is simplified with the cast to `UINT`, as it reduces the branches in the method.

### DIFF
--- a/src/bitmap.c
+++ b/src/bitmap.c
@@ -2013,7 +2013,7 @@ GdipBitmapSetPixel (GpBitmap *bitmap, INT x, INT y, ARGB color)
 	
 	data = bitmap->active_bitmap;
 
-	if ((x < 0) || (x > data->width) || (y < 0) || (y > data->height) || (data->reserved & GBD_LOCKED)) {
+	if ((UINT)x >= data->width || (UINT) y >= data->height || data->reserved & GBD_LOCKED) {
 		return InvalidParameter;
 	}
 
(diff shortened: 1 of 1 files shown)


filipnavara:
I, personally, find the previous code more readable and easier to follow. The compiler should be able to optimize it anyway. The change from `<` to`<=` looks correct though.

gfoidl:
> The compiler should be able to optimize it anyway

Unfortunately not. https://godbolt.org/z/oCLl0c shows a (contrieved) example. Neither gcc nor clang optimize this (with `-O3`).
That's also the reason why this was changed.

But I'll have no problem to revert this, as the previous version is more readable, I've to admit.

filipnavara:
I think there could be a reason why it doesn't optimize it. Checking against zero doesn't involve memory access to the `Bitmap` structure. It would make sense to put both of the `x < 0` and `y < 0` checks at the beginning to take advantage of that which generates one less branch. I think that's still quite readable.

gfoidl:
I don't know the reason either why this won't be optimized. But even [without memory access](https://godbolt.org/z/XGDc0S) gcc won't optimize this. clang does a bit better, but it's not optimized dows to a `jbe` instruction.

> Checking against zero doesn't involve memory access to the `Bitmap` structure

That's a good point. 
But under the assumption that `< 0` will be the exceptional code-path, the memory access will still happen -- once for checking width / height, second for accessing the data.
So we would optimize for the exceptional path instead for the the frequent / common path.

With the [proposed change](https://godbolt.org/z/lPgj8F) gcc doesn't optimize this. clang does a bit better with an `or`-trick.

Personally I would stick with the `UINT`-cast, as it is reduces the compares (and I'm used to it). But if you want, I'll update to `x < 0 || y < 0 || x >= width || y >= height` pattern.

filipnavara:
I find `x < 0 || y < 0 || x >= width || y >= height` much easier to read (personal opinion).

I don't see the value in micro-optimizing this since it's something that the branch predictor can predict quite reliably. Even if used in a tight loop it would probably have no performance impact unless you intentionally pass invalid values.

gfoidl:
I'll update to this.

Besides: the branch predictor will do a good job, but there's still the cost for the `cmp`. Anyway, that's micro-opts and I agree with you here.

directhex:

directhex:

PR #549, opened 2019-05-15
Add support for utf16 high/low surrogates in measurestring/drawstring.

bmerlet:
This patch allows Garphics.MeasureString() and Graphics.DrawString() to correctly measure and draw unicode characters with a codepoint greater than 0xffff.

### DIFF
--- a/src/font.c
+++ b/src/font.c
@@ -222,7 +222,7 @@ GdipPrivateAddFontFile (GpFontCollection *fontCollection, GDIPCONST WCHAR *filen
 	if (!fontCollection || !filename)
 		return InvalidParameter;
     
-	file = (BYTE*) ucs2_to_utf8 ((const gunichar2 *)filename, -1);
+	file = (BYTE*) utf16_to_utf8 ((const gunichar2 *)filename, -1);
 	if (!file)
 		return OutOfMemory;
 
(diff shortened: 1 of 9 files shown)


filipnavara:
The change looks ok, but the name of the function (ucs2_to_utf8) is a bit misleading now since the difference between `ucs2` and `utf16` is the introduction of surrogate pairs.

bmerlet:
Is this a request for me to change the name of the function?
I don't mind doing it, I am just not sure how to interpret the comment.

filipnavara:
Sorry, I should have been clearer :) I'd probably change the name to make it more obvious what the function does.

bmerlet:
So I renamed the function (to utf16_to_utf8()). I also fixed utf8_length_for_uscs2_string() to add the correct number of bytes for utf16 surrogates, and renamed it as well. Finally, I looked through the code for other uses of the 0xe000 and 0xd800 constants, found another one. They are used by utf8_to_ucs2(). This function is only used to translate font names from utf8 back to ucs2. I decided to leave utf8_to_ucs2() alone. Let me know if you think it ought to be upgraded as well.

bmerlet:
When are the Jenkins build supposed to run? It's been 5 days since this patch got approved...

filipnavara:
You need to be on a whitelist for Jenkins to run. /cc @akoeplinger

directhex:

directhex:

PR #558, opened 2019-06-21
GdipCreateBitmapFromScan0: Fix a buffer overflow with pixel formats < 32 bpp

qmfrederik:
If you create a new bitmap from scratch (that is, `scan0` is `NULL`), `GdipCreateBitmapFromScan0` will allocate memory to hold the raw bitmap data for you.

It then attempts to initialize this. By default, it will `memset` all bytes to 0, so you should get a black image in most formats. However, for formats with alpha-channels, this would set alpha to `0`, causing the image to be "completely translucent black" (AKA white).

So the method goes on a quest to set the alpha channel to 1. It iterates over the framebuffer using a 32-bit `ARGB` value. For pixel formats with < 32 bpp, this would cause a buffer overflow for the last pixel.

Fix the check for the bpp, and make sure we really have a 32bpp pixel format.

### DIFF
--- a/src/bitmap.c
+++ b/src/bitmap.c
@@ -902,14 +902,14 @@ GdipCreateBitmapFromScan0 (int width, int height, int stride, PixelFormat format
 			goto fail;
 		}
 
-		if ((gdip_get_pixel_format_bpp(format) < 16) || gdip_is_an_alpha_pixelformat(format)) {
+		if ((gdip_get_pixel_format_bpp(format) < 32) || gdip_is_an_alpha_pixelformat(format)) {
 			memset (scan0, 0, stride * height);
 		} else {
 			/* Since the pixel format is not an alpha pixel format (i.e., it is
-			 * either PixelFormat24bppRGB or PixelFormat32bppRGB), the image should be
-			 * initially black, not initially transparent. Thus, we need to set
-			 * the alpha channel, which the user code doesn't think exists but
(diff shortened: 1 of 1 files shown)


qmfrederik:
The code comment for `PixelFormat24bppRGB` really doesn't make sense to me - that format doesn't have an alpha channel (and nor does the underlying cairo format), so there's nothing that needs to be done in the first place.

qmfrederik:
Found this bug when running `GraphicsTest.FromImage_Bitmap_Success` from the CoreFX test suite, by the way.

src/bitmap.c
```diff
@@ -902,14 +902,14 @@ GdipCreateBitmapFromScan0 (int width, int height, int stride, PixelFormat format
 			goto fail;
 		}
 
-		if ((gdip_get_pixel_format_bpp(format) < 16) || gdip_is_an_alpha_pixelformat(format)) {
+		if ((gdip_get_pixel_format_bpp(format) < 32) || gdip_is_an_alpha_pixelformat(format)) {
```
filipnavara: Looks like the code should handle *non-alpha* formats where the space for alpha channel is unused (but Cairo cares about it). Looking at the list of pixel formats the obvious ones are PixelFormat32bppRGB and PixelFormat16bppRGB555. We never really handled the 16-bit case here and it may still be broken. The `if` now basically boils down to `format != PixelFormat32bppRGB`.

PR #560, opened 2019-06-21
Implement deep cloning of ImageAttributes

qmfrederik:
`GdipCloneImageAttributes` would do a `memcpy` of the `ImageAttributes` class, but did not do deep cloning of reference values. This leads to double frees (and other erratic behavior) when disposing the original `GpImageAttributes` and its clone.


qmfrederik:
`ImageAttributesTest.Clone_Success` in the corefx test suite excercises this code path.

filipnavara:

PR #561, opened 2019-06-21
Keep GraphicsContainer/GraphicsState non-zero

qmfrederik:
The current implementation of `GdipSaveGraphics` would set `GraphicsState` to the zero-based index of the current state in an array. The Windows implementation of System.Drawing [asserts that `GraphicsState` is non-zero](https://github.com/dotnet/corefx/blob/master/src/System.Drawing.Common/src/System/Drawing/Graphics.Windows.cs#L2690).

So simply increment `GraphicsState` by one when setting it, and substract one when reading it.


filipnavara:

PR #562, opened 2019-06-24
Matrix: Handle boundary values

qmfrederik:
GDI+ seems to map values between max(float) and infinity to max(float), instead of infinity. This is covered by a couple of tests in the corefx `MatrixTests` test suite.

This PR updates libgdiplus so that it matches the observed GDI+ behavior.


qmfrederik:
With this change, all `MatrixTests` pass, with the exception of a couple of rows in `Rotate_Matrix_Success`.

PR #565, opened 2019-06-24
CustomLineCap: Better argument validation

qmfrederik:
Only a couple of `LineCap` values are accepted by GDI+. All other values are ignored, and silently reset to `LineCapFlat`.

This behavior is covered by the `CustomLineCapTests` tests in corefx.


filipnavara:
The tests for customlinecaps failed, that has to be investigated.

PR #567, opened 2019-06-25
Various Image Attribute fixes

qmfrederik:
- In `gdip_process_bitmap_attributes`, check for `ImageAttributeFlagsNoOp`, and do not process any attributes if no-op is enabled
- Implement gamma correction
- Implement treshold correction
- Correctly implement the clearing of flags (`imgattr->flags &= ~ImageAttributeFlagsX` instead of `imgattr->flags |= ~ImageAttributeFlagsX`)
- Prototype CMYK channel support

### DIFF
--- a/src/imageattributes.c
+++ b/src/imageattributes.c
@@ -20,6 +20,7 @@
  * Authors:
  *	Jordi Mas i Hernandez <jordi@ximian.com>, 2004-2005
  *	Sebastien Pouliot  <sebastien@ximian.com>
+ *	Frederik Carlier  <frederik.carlier@quamotion.mobi>
  *
  */
 
@@ -92,7 +93,7 @@ gdip_process_bitmap_attributes (GpBitmap *bitmap, void **dest, GpImageAttributes
 {
 	GpStatus status;
 	GpImageAttribute *imgattr, *def;
-	GpImageAttribute *colormap, *gamma, *trans, *cmatrix, *treshold;
+	GpImageAttribute *colormap, *gamma, *trans, *cmatrix, *treshold, *cmyk;
 	GpBitmap *bmpdest;
 	ARGB color;
 	BYTE *color_p = (BYTE*) &color;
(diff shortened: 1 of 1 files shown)


src/imageattributes.c
```diff
@@ -204,23 +218,98 @@ gdip_process_bitmap_attributes (GpBitmap *bitmap, void **dest, GpImageAttributes
...
+
+				color = ((guint32)a << 24) | (r << 16) | (g << 8) | b;
+				GdipBitmapSetPixel(bmpdest, x, y, color);
+			}
+		}
+	}
+
 	/* Apply transparency range */
-	if (trans->flags & ImageAttributeFlagsColorKeysEnabled) {
+	/* FIXME: This will convert pixels which have a color in [key_colorlow, key_colorhey] to transparent pixels.
```
qmfrederik: @filipnavara Any thoughts on how ti fix this? The code correctly sets the pixels to be transparent, but `cairo_fill` will not copy transparent pixels, effectively rendering this a no-op.

filipnavara: @qmfrederik I briefly looked at it but it will require me to dive deeper into the code. I will review the remaining patches and try to get back to this one next week.

filipnavara:
Looks ok to me aside from the compilation errors (-Werror). That has to be fixed before merging.

src/imageattributes.c
```diff
@@ -129,8 +136,19 @@ gdip_process_bitmap_attributes (GpBitmap *bitmap, void **dest, GpImageAttributes
...
-	if ((colormap->flags & ImageAttributeFlagsColorRemapTableEnabled) || (gamma->flags & ImageAttributeFlagsGammaEnabled) || (trans->flags & ImageAttributeFlagsColorKeysEnabled) ||
-	    ((cmatrix->flags & ImageAttributeFlagsColorMatrixEnabled) && cmatrix->colormatrix != NULL)) {
+	if (imgattr->flags & ImageAttributeFlagsOutputChannelEnabled) {
+		cmyk = imgattr;
+	}
+	else {
+		cmyk = def;
+	}
+
+	if (!(colormap->flags & ImageAttributeFlagsNoOp) && (colormap->flags & ImageAttributeFlagsColorRemapTableEnabled)
```
directhex: `imageattributes.c:146:51: error: suggest parentheses around '&&' within '||' [-Werror=parentheses]`

src/imageattributes.c
```diff
@@ -204,23 +218,100 @@ gdip_process_bitmap_attributes (GpBitmap *bitmap, void **dest, GpImageAttributes
...
+	          However, GdipDrawImageRectRect in image.c uses cairo_fill to draw these transparent pixels onto
+			  the target surface. This method will not copy the transparency value; rather, it will ignore
+			  transparent values. */
+	if (!(trans->flags & ImageAttributeFlagsNoOp) && trans->flags & ImageAttributeFlagsColorKeysEnabled) {
 		for (int y = 0; y < bitmap->active_bitmap->height; y++) {
 			for (int x = 0; x < bitmap->active_bitmap->width; x++) {
 
 				GdipBitmapGetPixel (bmpdest, x, y, &color);
-
+				color &= ~ALPHA_MASK;
```
filipnavara, line 304: @qmfrederik Do you remember what was the rationale for this changed line?

It breaks some user code (https://github.com/mono/mono/issues/8514) because it masks only the input value and not the values in `trans->key_colorlow` and `trans->key_colorhigh`. Thus the condition for transparent pixels is not matched. It should either mask on both places or neither of them.

filipnavara: Wine compares just the R, G, B values do I'd guess that the correct way is to mask both.

PR #569, opened 2019-06-26
font-config: Use a rudimentary default configuration when no config file can be found

qmfrederik:
fontconfig will not enumerate _any_ fonts if it cannot find a configuration file.
This PR tries to set some sensible defaults in that scenario, so that at least the system-wide fonts can be found.

### DIFF
--- a/src/general.c
+++ b/src/general.c
@@ -64,7 +64,49 @@ GdiplusStartup (ULONG_PTR *token, const GdiplusStartupInput *input, GdiplusStart
 	if (status != Ok)
 		return status;
 
-	FcInit ();
+	FcInit();
+
+	/* A fontconfig instance which didn't find a configfile is unbelievably
+		cranky, so let's just write out a small xml file and make fontconfig
+		happy */
+	if (FcConfigFilename(0) == NULL)
+	{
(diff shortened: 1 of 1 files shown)


src/general.c
```diff
@@ -64,7 +64,49 @@ GdiplusStartup (ULONG_PTR *token, const GdiplusStartupInput *input, GdiplusStart
...
+	/* A fontconfig instance which didn't find a configfile is unbelievably
+		cranky, so let's just write out a small xml file and make fontconfig
+		happy */
+	if (FcConfigFilename(0) == NULL)
+	{
+		/* Newer versions of font-config have FcConfigParseAndLoadFromMemory,
+		   which we could use to avoid generating a temporary file. But meanwhile,
+		   we are stuck with this workaround. */
+		char namebuf[512];
+		tmpnam(namebuf);
```
directhex: This causes a build failure on Travis (which uses --enable-warnaserror):
```
general.c:78:9: error: ignoring return value of ‘tmpnam’, declared with attribute warn_unused_result [-Werror=unused-result]
752   tmpnam(namebuf);
```

PR #571, opened 2019-07-30
fix libm not getting used when building

NattyNarwhal:
this is needed for AIX/PASE


directhex:
The failures from Travis are unrelated. Merging.

PR #574, opened 2019-08-01
Bump minimum version of Pango and add C version of MeasureString_Wrapping_Dots test

filipnavara:


tests/testtext.c
```diff
@@ -46,6 +46,7 @@ static void test_measure_string(void)
 	const WCHAR teststring1[] = { 'M', '\n', '\n', 'M', 0 };
```
jaykrell: These should all be `static`. Some compilers will do that. Visual C++ will copy them to the stack.

filipnavara: Well, I just followed the code that was already there. We can fix this issue separately.

filipnavara:
This is one funky bug... It works fine on macOS with Pango 1.42 and it doesn't work on Ubuntu 16.04 with Pango 1.38.1. The problem is that on Ubuntu the "." characters are not marked with "char_break" logical attribute so the Pango layout fails to insert a line break. On macOS all the characters have the "char_break" logical attribute and it returns the correct result.

filipnavara:
The fix in Pango was implemented in https://github.com/GNOME/pango/commit/e8316c1c6f46b844b265478622d3a581305b70b6#diff-3d8168315b940c3e9d29a4991735a9a7 so any version newer than 1.40.14 is fine. I'll bump the minimum requirement.

PR #577, opened 2019-08-02
Fix a couple static analyzer warnings

akoeplinger:
Various instances of "Multiplication result may overflow 'unsigned int' before it is converted to 'size_t'." and "comparison is always the same" warnings.

### DIFF
--- a/src/graphics.c
+++ b/src/graphics.c
@@ -1538,7 +1538,7 @@ GdipFillClosedCurve2 (GpGraphics *graphics, GpBrush *brush, GDIPCONST GpPointF *
 	if (tension == 0)
 		return GdipFillPolygon2 (graphics, brush, points, count);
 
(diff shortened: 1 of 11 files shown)


src/graphics.c
```diff
@@ -1538,7 +1538,7 @@ GdipFillClosedCurve2 (GpGraphics *graphics, GpBrush *brush, GDIPCONST GpPointF *
 	if (tension == 0)
 		return GdipFillPolygon2 (graphics, brush, points, count);
 
-	if (!graphics || !brush || !points || count <= 0)
```
hughbe: Surely this changes behaviour if count==0

akoeplinger: @hughbe no because a few lines above we return from the function `if (count < 3)`

PR #578, opened 2019-08-02
Change to 6.0 for release. So many changes deserves a big number

directhex:
Let's get the static analyser stuff in, then this, and tag, and stuff


directhex:
Waiting on https://github.com/mono/libgdiplus/pull/577 before we do this thing

configure.ac
```diff
@@ -1,4 +1,4 @@
-AC_INIT(libgdiplus, [5.8],
+AC_INIT(libgdiplus, [6.0],
 	[https://bugzilla.xamarin.com/enter_bug.cgi?product=Class%20Libraries&component=libgdiplus])
```
akoeplinger: lol, bugzilla. will fix it.

PR #581, opened 2019-08-04
Fix warnings running autogen.sh

hughbe:
```In file included from jpegcodec.c:40:
In file included from /usr/local/include/jpeglib.h:27:
/usr/local/include/jmorecfg.h:199:24: warning: redefinition of typedef 'UINT16'
      is a C11 feature [-Wtypedef-redefinition]
typedef unsigned short UINT16;
                       ^
./win32structs.h:347:17: note: previous definition is here
typedef guint16 UINT16;
```

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -131,7 +131,7 @@ do
   shift
 done
 
-conf_flags="--enable-maintainer-mode --enable-compile-warnings" #--enable-iso-c
+conf_flags="--enable-maintainer-mode"
 
 if test x$NOCONFIGURE = x; then
   echo Running $srcdir/configure $conf_flags $CONF_OPTIONS ...
--- a/configure.ac
+++ b/configure.ac
@@ -31,12 +31,12 @@ PKG_CHECK_MODULES(CAIRO, cairo >= $CAIRO_REQUIRED_VERSION)
 # Optional use of Pango's text rendering on top of Cairo
 AC_ARG_WITH(pango, [  --with-pango            use Pango to measure and draw text],[text_v=pango],[text_v=default])
 
-GDIPLUS_CFLAGS="$GDIPLUS_CFLAGS -Wall -Wextra -Wno-unused-parameter -Wno-sign-compare -std=gnu99"
+GDIPLUS_CFLAGS="$GDIPLUS_CFLAGS -Wall -Wextra -Wno-unused-parameter -Wno-sign-compare -std=gnu11"
 
 AC_ARG_ENABLE(warnaserror, AS_HELP_STRING([--enable-warnaserror],[Enable treating warnings as errors.]),[warnaserror=yes],[warnaserror=no])
 
 if test $warnaserror = "yes"; then
-  GDIPLUS_CFLAGS="$GDIPLUS_CFLAGS -Werror -Wno-unknown-warning-option -Wno-clobbered -Wno-typedef-redefinition"
+  GDIPLUS_CFLAGS="$GDIPLUS_CFLAGS -Werror -Wno-unknown-warning-option -Wno-clobbered"
 fi
 
 AC_ARG_ENABLE(asan, AS_HELP_STRING([--enable-asan],[Enable address sanitizer.]),[asan=yes],[asan=no])
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -4,7 +4,7 @@ LIBS = $(GDIPLUS_LIBS)
 
 AM_CPPFLAGS =					\
 	-I$(top_srcdir)				\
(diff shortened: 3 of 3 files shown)


tests/Makefile.am
```diff
@@ -352,19 +368,3 @@ TESTS = \
 if HAVE_X11
 # TESTS += testgdi
 endif
-
-EXTRA_DIST = \
```
hughbe: > tests/Makefile.am:356: warning: EXTRA_DIST multiply defined in condition TRUE ...
tests/Makefile.am:273: ... 'EXTRA_DIST' previously defined here

autogen.sh
```diff
@@ -131,7 +131,7 @@ do
   shift
 done
 
-conf_flags="--enable-maintainer-mode --enable-compile-warnings" #--enable-iso-c
+conf_flags="--enable-maintainer-mode" #--enable-iso-c
```
hughbe: > Running ./configure --enable-maintainer-mode --enable-compile-warnings --without-x11 --enable-asan ...
configure: WARNING: unrecognized options: --enable-compile-warnings

Should I also remove `#-enable-iso-c`

configure.ac
```diff
@@ -31,12 +31,12 @@ PKG_CHECK_MODULES(CAIRO, cairo >= $CAIRO_REQUIRED_VERSION)
 # Optional use of Pango's text rendering on top of Cairo
 AC_ARG_WITH(pango, [  --with-pango            use Pango to measure and draw text],[text_v=pango],[text_v=default])
 
-GDIPLUS_CFLAGS="$GDIPLUS_CFLAGS -Wall -Wextra -Wno-unused-parameter -Wno-sign-compare -std=gnu99"
+GDIPLUS_CFLAGS="$GDIPLUS_CFLAGS -Wall -Wextra -Wno-unused-parameter -Wno-sign-compare -std=gnu11"
```
akoeplinger: I have a hunch we might not have a C11 compiler everywhere ...

Can you please try whether something like this fixes the warning too?

```c
#ifndef UINT16
typedef guint16 UINT16;
#endif
```

configure.ac
```diff
@@ -252,7 +252,7 @@ AC_ARG_WITH([libjpeg],
 	[libjpeg_prefix="$withval"], [libjpeg_prefix="${prefix}"])
 
   if test x$with_libjpeg != xno && test -z "$LIBJPEG"; then
-    JPEG_CFLAGS="-I${libjpeg_prefix}/include"
+    JPEG_CFLAGS="-Isystem${libjpeg_prefix}/include"
```
akoeplinger: what's the reason for this change?

hughbe: Experiment - reverted

configure.ac
hughbe: Does not work, sadly

hughbe: > I have a hunch we might not have a C11 compiler everywhere ...

Are we not just then lying because the code shouldn't currently comply with a compiler that only supports c99!

akoeplinger: yeah but e.g. CentOS 6 ships with an ancient GCC which doesn't support C11 and it was not an issue there.

I just learned though that we build with a newer GCC on that old distro anyway so this is probably fine.

configure.ac
```diff
@@ -31,12 +31,12 @@ PKG_CHECK_MODULES(CAIRO, cairo >= $CAIRO_REQUIRED_VERSION)
 # Optional use of Pango's text rendering on top of Cairo
 AC_ARG_WITH(pango, [  --with-pango            use Pango to measure and draw text],[text_v=pango],[text_v=default])
 
-GDIPLUS_CFLAGS="$GDIPLUS_CFLAGS -Wall -Wextra -Wno-unused-parameter -Wno-sign-compare -std=gnu99"
+GDIPLUS_CFLAGS="$GDIPLUS_CFLAGS -Wall -Wextra -Wno-unused-parameter -Wno-sign-compare -Wno-typedef-redefinition -std=gnu99"
```
hughbe: @akoeplinger I've resolved to fix this by moving to unconditionally prevent this warning

akoeplinger: ok that works too but as I mentioned in the other thread we should be fine to bump to use C11 :)

tests/Makefile.am
```diff
@@ -306,7 +306,22 @@ EXTRA_DIST =			\
...
+	test.gif \
+	test.ico \
+	test.jpg \
+	test.otf \
+	test.png \
+	test.tif \
+	test.ttf \
+	test.wmf \
+	testhelpers.h \
+	$(NULL)
```
akoeplinger: any reason for this?

PR #587, opened 2019-08-04
Add image tests and fix some compat bugs

hughbe:


src/bitmap.c
```diff
@@ -2071,10 +2071,10 @@ GdipBitmapGetPixel (GpBitmap *bitmap, INT x, INT y, ARGB *color)
 
 	data = bitmap->active_bitmap;
 
-	if (x < 0 || y < 0 || x >= data->width || y >= data->height || data->reserved & GBD_LOCKED)
-		return InvalidParameter;
-
 	if (gdip_is_an_indexed_pixelformat (data->pixel_format)) {
+		if (x < 0 || x >= data->width || y < 0 || y >= data->height)
```
akoeplinger: do we not need to check `data->reserved & GBD_LOCKED` here?

hughbe: Tests show that we do not

PR #588, opened 2019-08-04
Fix leak in GdiplusStartup

hughbe:


tests/Makefile.am
```diff
@@ -306,7 +306,9 @@ EXTRA_DIST =			\
 	$(testtexturebrush_SOURCES)	\
 	$(testtext_SOURCES)	\
 	$(testtiffcodec_SOURCES) \
-	$(testwmfcodec_SOURCES)
+	$(testwmfcodec_SOURCES) \
+	lsansuppressions.txt \
```
akoeplinger: this file is missing

tests/Makefile.am
```diff
@@ -306,7 +306,9 @@ EXTRA_DIST =			\
...
+	$(NULL)
```
akoeplinger: what is this?

README.md
```diff
@@ -60,6 +60,14 @@ To run the tests with Clang sanitizers, run the following command from the root
 	./autogen.sh --enable-asan
 	make check
 
+To run the unit tests with leak sanitizers, run the following command from the root of the repository:
+
+	./autogen.sh --enable-asan
+	export ASAN_OPTIONS=detect_leaks=1
+	export ASAN_OPTIONS=fast_unwind_on_malloc=0
```
akoeplinger: @hughbe do we need this when we compile with ` -fno-omit-frame-pointer`? according to the FAQ (https://github.com/google/sanitizers/wiki/AddressSanitizer#faq) one of them should be enough

hughbe: On my linux VM things only worked when both were set

filipnavara: How does the second `export` not override the value from the first one?

hughbe: it doesn't seem to do that on my computer!

akoeplinger: Ok well it doesn't hurt since it's just in the docs :)

PR #594, opened 2019-08-10
JPEG codec: initialize cinfo.mem

qmfrederik:
If `image->active_bitmap->pixel_format` is not a valid pixel format, the call to `goto error` would cause a call to `jpeg_destroy_compress` which would access `cinfo.mem`. Initialize it to `NULL` to avoid an AV.


akoeplinger:
@qmfrederik I added you to the repo with `Read` permissions, that allows you to use the `/azp` commands (weird Azure Pipelines limitation).

PR #595, opened 2019-08-10
Use pango 1.42.4

qmfrederik:
`pango_fc_font_unlock_face` is depcreated in Pango 1.44, which got pulled into the Travis build via a Homebrew update.

This PR pins pango to 1.42 to resolve that issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ addons:
     - git
 
 install:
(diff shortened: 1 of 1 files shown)


akoeplinger:
Can we fix the deprecation issue instead of pinning to 1.42? According to the docs:

```
pango_fc_font_unlock_face has been deprecated since version 1.44 and should not be used in newly-written code.

Use pango_font_get_hb_font() instead.
```

filipnavara:
Can we close this now? Different patch for the same issue landed in c965228fc3d1ad9dafb0330ba384aeb7b770b301. /cc @akoeplinger

qmfrederik:
Yes :)

PR #596, opened 2019-08-12
Read the X and Y pixel density from EXIF data

### LINKED ISSUE #539: For some images HorizontalResolution and VerticalResolution erroneusly returns 0
Porting over https://bugzilla.xamarin.com/show_bug.cgi?id=33310

I just reproduced this for some images on my local using version 5.6

qmfrederik:
Fetch the X and Y pixel density from the EXIF data if the they could not be retrieved from the JPEG metadata.

Fixes #539

### DIFF
--- a/src/jpegcodec.c
+++ b/src/jpegcodec.c
@@ -556,30 +556,30 @@ add_properties_from_content (ExifContent *content, void *user_data)
 }
 
 static int
-get_exif_tag_value_as_int(ExifData* exif, ExifTag tag)
+get_exif_tag_value_as_int (ExifData* exif, ExifTag tag)
 {
-	ExifEntry* entry = exif_data_get_entry(exif, tag);
+	ExifEntry* entry = exif_data_get_entry (exif, tag);
 
 	if (entry) {
 		ExifByteOrder order;
 
-		order = exif_data_get_byte_order(exif);
-		return (int)exif_get_short(entry->data, order);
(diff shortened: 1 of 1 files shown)


qmfrederik:
Here's the unit test I used, with the image from https://xamarin.github.io/bugzilla-archives/33/33310/bug.html

```
TEST(JpegCodecTest, GetVerticalResolution) {
	STARTUP

	GpBitmap* bitmap = NULL;
	ASSERT_EQ(Ok, GdipCreateBitmapFromFile(L"00002.jpg", &bitmap));

	REAL resolution;
	ASSERT_EQ(Ok, GdipGetImageVerticalResolution(bitmap, &resolution));
	ASSERT_EQ(180, resolution);

	ASSERT_EQ(Ok, GdipGetImageHorizontalResolution(bitmap, &resolution));
	ASSERT_EQ(180, resolution);

	SHUTDOWN
}
```

src/jpegcodec.c
```diff
@@ -555,6 +555,36 @@ add_properties_from_content (ExifContent *content, void *user_data)
 	exif_content_foreach_entry (content, add_properties_from_entry, user_data);
 }
 
+static int
+get_exif_tag_value_as_int(ExifData* exif, ExifTag tag)
```
akoeplinger: code style: space before `(`.

Applies to the whole PR 😄

PR #602, opened 2019-10-02
[RFC] Use clang-format code style

akoeplinger:
What do you guys think about having a common code style using clang-format?

This is a first draft, if you spot some problematic pattern we can check whether we can tweak the settings.

@filipnavara @qmfrederik @hughbe


src/gdipluspixelformats.h
```diff
@@ -36,96 +36,96 @@ typedef DWORDLONG ARGB64;
...
-	PaletteTypeFixedBW = 2,
-	PaletteTypeFixedHalftone8 = 3,
-	PaletteTypeFixedHalftone27 = 4,
-	PaletteTypeFixedHalftone64 = 5,
+	PaletteTypeCustom	   = 0,
+	PaletteTypeOptimal	  = 1,
+	PaletteTypeFixedBW	  = 2,
+	PaletteTypeFixedHalftone8   = 3,
+	PaletteTypeFixedHalftone27  = 4,
+	PaletteTypeFixedHalftone64  = 5,
```
filipnavara: This looks oddly unaligned.

filipnavara:
In general I am fine with whatever style you prefer as long as VS Code can enforce it for me. It seems that there is an extension for that (https://marketplace.visualstudio.com/items?itemName=xaver.clang-format) but I didn't try it yet.

qmfrederik:
Same here. I'm all about enforcing a common style, less opinionated about which style that should be :)

hughbe:
Happy with this

PR #605, opened 2019-10-03
Reinstate support for Pango 1.44+

filipnavara:


akoeplinger:
I'm getting this native crash when running with pango 1.44 on mono-netcore on macOS: https://gist.github.com/akoeplinger/156782f7e7063e028d4c7f078663c965

filipnavara:
I'll check it.

filipnavara:
It didn't crash on my machine yet. I do however get some errors:

```
dotnet(22556,0x70000de58000) malloc: can't allocate region
*** mach_vm_map(size=18446744071562067968) failed (error code=3)
dotnet(22556,0x70000de58000) malloc: *** set a breakpoint in malloc_error_break to debug
    System.Drawing.Tests.FontFamilyTests.Families_Get_ReturnsExpected [FAIL]
      Assert.Equal() Failure
                ↓ (pos 0)
      Expected: Noto Sans Inscriptional Parthia
      Actual:   Verdana
                ↑ (pos 0)
      Stack Trace:
        /_/src/System.Drawing.Common/tests/FontFamilyTests.cs(126,0): at System.Drawing.Tests.FontFamilyTests.Families_Get_ReturnsExpected()
        /Users/filipnavara/Documents/mono/netcore/System.Private.CoreLib/src/System.Reflection/RuntimeMethodInfo.cs(339,0): at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
```

akoeplinger:
Let's see if it crashes in CI.

/azp run

filipnavara:
It works on the CI but I still want to look into the failures locally.

akoeplinger:
Turns out my issue was having XQuartz installed and libgdiplus built against it, which causes mismatches between cairo from X11 and the one from Homebrew.

filipnavara:
The "Noto Sans Inscriptional Parthia" is orthogonal to this PR. I still plan to look into it but it should not be a blocker.

PR #612, opened 2019-11-18
Add unit test for reading X and Y pixel density from EXIF data

### LINKED ISSUE #611: Non-JFIF JPEG image has resolution 0/0 on Linux
I noticed that .NET Core 3 reports the resolution of some of my JPEG images as 0/0 on Linux but not on Windows. These images are EXIF/JPEG rather than JFIF/JPEG.

Git repo that shows the problem: https://github.com/provegard/jpegtest

On Windows, the output of `dotnet run` is the expected 300/300, but on Linux it's 0/0.

I tested on Debian with libgdiplus version 4.2-2.

I chose to open the issue here rather than over at [corefx](https://github.com/dotnet/corefx) because it appears to me that the .NET Core core just delegates to libgdiplus. If that's wrong, please let me know and I'll move the issue.

provegard:
See 8c5b0d48317af4c6fd702a1372f58a487e934f67

Test cases added as part of trying to fix #611 before I realized that the current fix works, it's just not part of any released version.

I have verified that the test case fails correctly without the code added in 8c5b0d48317af4c6fd702a1372f58a487e934f67.


provegard:
By the way, the test image is not mine, it's this one: 

https://github.com/eugeneware/jpeg-js/blob/master/test/fixtures/tree-rgb.jpg

provegard:
Please let me know if you think I should amend the PR with image source/licence information to better comply with the CLA.

akoeplinger:
Yeah I'm not sure we can take that file as is so I created a variation of one of our existing test images and use it instead :)

PR #613, opened 2019-11-20
Fixing Resource Leaks identified by Coverity Scan

n432sean:
These Resource Leaks were identified at https://scan.coverity.com/projects/libgdiplus

### DIFF
--- a/src/bmpcodec.c
+++ b/src/bmpcodec.c
@@ -1197,6 +1197,7 @@ gdip_read_bmp_image (void *pointer, GpImage **image, ImageSource source)
 			status = gdip_read_bmp_indexed (pointer, pixels, upsidedown, originalFormat, originalStride, result->active_bitmap->width, result->active_bitmap->height, source);
 			if (status != Ok) {
 				gdip_bitmap_dispose (result);
+				GdipFree(pixels);
 				return status;
 			}
 		}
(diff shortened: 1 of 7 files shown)


Tampa:
This might just be what it found, but there could be more, the latest release is causing massive memory leaks for various mono apps in some cases a couple GB per hour not cleared. Solution? Revert back to distro version which in some cases is 2 versions behind current. This hopefully fixes some of the leaks, it should really be merged asap!

akoeplinger:
@Tampa do you have some sample code/project which reproduces the leaks you're seing?

akoeplinger:
I cherry-picked this commit to master as well.

Tampa:
OpenSimulator with mono 6.4 and libgdiplus 6.0.4 was 4gb average usage, now within a day 34gb. You can find me on their freenode irc #opensim-dev, setup of that project is a bit difficult would spam this comment box to next week.

akoeplinger:
Hm that sounds like a quite complicated setup to reproduce the issue, do you think you could create some smaller sample? I don't have a lot of time right now unfortunately.

src/bmpcodec.c
```diff
@@ -1197,13 +1197,15 @@ gdip_read_bmp_image (void *pointer, GpImage **image, ImageSource source)
...
 				gdip_bitmap_dispose (result);
+				GdipFree(pixels);
 				return status;
 			}
 		}
 	} else {
 		status = gdip_read_bmp_scans (pointer, pixels, upsidedown, originalFormat, originalStride, result->active_bitmap->stride, result->active_bitmap->width, result->active_bitmap->height, source);
 		if (status != Ok) {
 			gdip_bitmap_dispose (result);
+			GdipFree(pixels);
```
akoeplinger: FWIW I had to revert these two changes as they caused the testbitmap test to crash in master.

Tampa:
I could try to build master and see if I can replace the installed version with that and do testing, if there was a repo I could just add to apt that would make it easier though(not to keen to building on cloud machines). If you need any debug or reports let me know how to generate them, I have a fairly singular testing method to reproduce the leak now.

Issue #620, opened 2020-01-12
gtest is broken on CentOS 7


NattyNarwhal:
There's a [CMake 2.8 incompatibility](google/googletest#1786) with the current commit. Merging google/googletest#2242's commit solves this; at least on ppc64.

PR #621, opened 2020-01-13
Update googletest to v1.10

akoeplinger:
Fixes https://github.com/mono/libgdiplus/issues/620


xkszltl:
@akoeplinger 
Probably forgot about this?
https://github.com/mono/libgdiplus/blob/94a49875487e296376f209fe64b921c6020f74c0/.gitmodules#L4

Issue #627, opened 2020-02-26
Characters are missing when using Japanese on Windows Forms


s520:
## Steps to Reproduce

1. Install the Japanese language related packages of the distribution.

2. Change the system character set to Japanese.

   ```shell
   # localectl set-locale LANG=ja_JP.UTF-8
   # source /etc/locale.conf
   ```

3. Execute the program using Windows Forms with Japanese displayed. Ex) [winforms-test.zip](https://github.com/mono/libgdiplus/files/4254346/winforms-test.zip)

### Current Behavior

Some characters are missing.

- mono 5.18.1.3 + libgdiplus 6.0.2

  ![linux_libgdiplus-6 0 2](https://user-images.githubusercontent.com/29241703/75327138-376b5400-58bf-11ea-826d-bb6e903feb23.png)

  The text on the right button has disappeared.

- mono 5.18.1.3 + libgdiplus 6.0.4

  ![linux_libgdiplus-6 0 4](https://user-images.githubusercontent.com/29241703/75327170-46520680-58bf-11ea-9692-baf37d52f364.png)

  The text on the right button is visible, but the text on the left label is missing.

### Expected Behavior

Characters are displayed normally as in Windows and Linux + libgdiplus 5.6.

- Windows

  ![windows](https://user-images.githubusercontent.com/29241703/75327221-55d14f80-58bf-11ea-8569-620c915cae6f.png)

- mono 5.18.1.3 + libgdiplus 5.6

  ![linux_libgdiplus-5 6](https://user-images.githubusercontent.com/29241703/75327237-5f5ab780-58bf-11ea-9d47-94cce5a9d3c1.png)

## On which platforms did you notice this

[ ] macOS
[X] Linux
[ ] Windows

**Version Used**:

mono: 6.8.0.105, libgdiplus: 6.0.4

## Postscript

Not only me but also my friends have similar symptoms.

I am using fedora 30 and he is using Ubuntu 18.04.

[He previously reported this problem to the mono repository.](https://github.com/mono/mono/issues/17824)

Later he found that downgrading libgdiplus to 4.2 works fine.

So I decided to find out what version worked.

I downgraded mono to 5.18.1.3 and upgraded libgdiplus from 5.6 to 6.0.4.

As a result, it was normal until 5.6.1, but a problem has occurred since 6.0.2.

In addition, this problem seems to occur even in Chinese.

s520:
I went back to commits.
It works fine until the next commit.
[...]

PR #628, opened 2020-02-26
Fix: an issue where Japanese characters are recognized as spaces

s520:
This PR fix an issue where Japanese characters are recognized as spaces. (#627)

This fix does not recur issue (https://github.com/mono/mono/issues/8272) fixed by #370.

### DIFF
--- a/src/text-cairo.c
+++ b/src/text-cairo.c
@@ -27,7 +27,7 @@
 
 #ifndef USE_PANGO_RENDERING
 
-#include <ctype.h>
+#include <wctype.h>
 
 #include "text-cairo-private.h"
 #include "graphics-private.h"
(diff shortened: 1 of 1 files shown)


qmfrederik:
@s520 Do you think you could include a unit test which covers your case and possible #8272? That way, we can make sure this doesn't accidentally regress next time somebody touches the code 😄 .

s520:
First of all, my problem does not occur unless the system language is Japanese.
In English environment, the characters are □ (white square), but the number of characters looks correct.
I need to check the behavior of function GdipMeasureString due to differences in system languages.

s520:
Well, it's hard for me to add this unit test.
Do you have a good idea?

qmfrederik:
@s520 I'll have a look.

Just to be clear, I don't have commit rights to this repository, so I don't decide what gets merged and what not.

Since there have been various fixes for this piece of code which have had side-effects, I'd be nice to have some unit tests. I'll see if there's an easy way to get some set up.

marcinm100:
Unfortunately the problem from #8272 has returned.

s520:
@marcinm100 
Does your problem occur under the same conditions as shown in https://github.com/mono/mono/issues/8272?
I have confirmed that the problem does not recur under the conditions shown in https://github.com/mono/mono/issues/8272.
Please indicate the conditions under which your problem occurs.

marcinm100:
Yes, my locale is set to en_US.UTF-8 and the problem occurs under the same conditions as in #8272.

Issue #626, opened 2020-02-13
Drawing (Images, Font) issues with mono 6.8.105 and libgdiplus 6.0.4


g0ldi:
## Steps to Reproduce

1. Mono version 6.8.105 is installed (pkg: mono-complete)
2. Compile this [sample app source](https://github.com/mono/mono/files/4180763/mono_drawingissue.zip) with `msbuild mono_6.8_drawingissue.csproj`
3. libgdiplus version 6.0.4 is installed (pkg: libgdiplus=6.0.4-0xamarin1+debian9b1)
4. Start the application `mono_drawingissue.exe` (see attached screenshots)
5. Downgrade libgdiplus to version 5.6 (pkg: libgdiplus=5.6-0xamarin5+debian9b1)
6. Start again the application `mono_drawingissue.exe` (see attached screenshots)
7. Notice the difference between the application even it's the same appliacation.

### Current Behavior
**Note:** Issue can be observed/reproduced with mono version **6.8.96** and **6.8.105**.

#### Mono 6.8.105 with libgdiplus 6.0.4-0xamarin1+debian9b1
With the packages specified in the header the 
- image specified for the picturebox is drawn but not visible
- text "Log On" is drawn but not visible
- text of buttons (LBtn, MLBtn, MRBtn, RBtn) are not drawn using the specified font Arial, bold.
- image and text "Log On" are redrawn and **are visible** when another window is moved over the form. 

##### Screenshots
- Application drawn when started (notice the button font):
![image](https://user-images.githubusercontent.com/9973579/74152393-843c0300-4c0e-11ea-9c2d-ad0fe4baa64f.png)

- Application drawn when moving partially another screen over it:
![image](https://user-images.githubusercontent.com/9973579/74152407-8736f380-4c0e-11ea-98a3-f0ba27246958.png)

#### Mono 6.8.105 with libgdiplus 5.6-0xamarin5+debian9b1
- image specified for the picturebox is drawn and is visible
- text "Log On" is drawn and is visible
- text of buttons (LBtn, MLBtn, MRBtn, RBtn) are drawn using the specified font Arial, bold.

##### Screenshots
- Application drawn when started (notice the button font):
![image](https://user-images.githubusercontent.com/9973579/74152418-8a31e400-4c0e-11ea-886c-d64a260e2913.png)

### Expected Behavior
[...]

PR #629, opened 2020-02-26
Bugfix/font style issue

abossard:
This fixes the wrong font-weight from issue #626 

Sadly I it breaks the tests, but it does render the font as Bold.

I'll check what's wrong with the tests. Is this maybe a Mono issue instead of a libgdiplus issue?

### DIFF
--- a/src/font.c
+++ b/src/font.c
@@ -890,7 +890,8 @@ gdip_get_cairo_font_face (GpFont *font)
 {
 	if (!font->cairofnt) {
 		FcPattern *pattern = FcPatternBuild (
-			FcPatternDuplicate (font->family->pattern),
+			NULL,
+			FC_FAMILY, FcTypeString,  font->face,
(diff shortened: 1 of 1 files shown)


abossard:
I reverted the test metrics (cell ascent and descent and letter spacing) to the previous values. I compiled a few examples and didn't notice a degradation.

filipnavara:
The fix in the current form is incorrect, as pointed out in issue #626. The reason for duplicating properties of the original font is to preserve path to custom fonts (`FC_FONT`, `FC_INDEX`). The font metrics in the test are from the custom font and if you observe different ones it means the font was incorrectly substituted.

engrch:
@filipnavara do you know which component is responsible for the Font substitution?

It looks like the FcPatternDuplicate does not copy the font correctly. Certain fonts in Windows are different in Linux; I wonder if the FcPatternDuplicate would handle such cases. (e.g. Arial in Linux is: Liberation Sans)

I built the latest mono from master and I can confirm that this revert of FcPatternDuplicate resolves the font issue for me. 

Debian with LightDM
Mono: Built from master
libgdiplus: master + this patch

PR #635, opened 2020-03-28
Fix multiple memory leaks for GpBitmap image data

filipnavara:
`gdip_bitmap_dispose` was leaking the bitmap backing data for certain image formats. This is essentially the same issue as in https://github.com/mono/libgdiplus/commit/1afe4711f10bd52e14a4c7985eae7414eda4108c but on a different place. I now unified both of the places to use the same code.

### DIFF
--- a/src/bitmap.c
+++ b/src/bitmap.c
@@ -590,6 +590,7 @@ gdip_bitmap_setactive(GpBitmap *bitmap, const GUID *dimension, int index)
 	}
 
 	/* Invalidate the cached surface */
+	gdip_bitmap_flush_surface (bitmap);
 	gdip_bitmap_invalidate_surface (bitmap);
 
 	if ((bitmap->num_of_frames == 0) || (bitmap->frames == NULL)) {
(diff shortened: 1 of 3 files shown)


immarvin:
hi @filipnavara @akoeplinger , what is the latest libgdi+ release including this PR and https://github.com/mono/libgdiplus/pull/603? if not, do you have any planned date to release the libgdi+ with the fix?
thanks
we are suffering from the memory leak issue

akoeplinger:
There's no release yet. @filipnavara do you think there are any leaks left you'd like to fix? I backported your latest ones to the release/6.0 branch already.

filipnavara:
@akoeplinger There are some leaks left in the test code and the issue with font collection leaking. I am not too concerned about the test code. The font collection leaking was there for many years and it's not a huge deal when consumed by the System.Drawing code. I guess for release/6.0 it's good to go. I'd also like to finish reviewing the last pending PRs and get 6.1 out soon after.

akoeplinger:
Ok sounds good, then I guess we can cut a 6.0 release soon.

n432sean:
Please consider https://github.com/mono/libgdiplus/pull/646 for the imminent release.

akoeplinger:
Ok, I tagged https://github.com/mono/libgdiplus/releases/tag/6.0.5

PR #637, opened 2020-03-31
WIP: Fix ownership of GpFontFamily objects returned by GdipGetFontCollectionFamilyList

filipnavara:
The returned objects are owned by GpFontCollection and the font collection is responsible for their lifetime. The managed implementation of System.Drawing in Mono and System.Drawing.Common in .NET Core already treated the ownership correctly which resulted in memory leaks. Change GdipDeleteFontFamily to detect accidental attempts to delete the font family objects.


filipnavara:
On second thought, the second commit should not be necessary. The failing asserts on CI were coming from finalizers of non-disposed objects. That still needs to be investigated.

PR #639, opened 2020-04-03
Keep clipping inside containers.

PreferLinux:
Containers need to maintain the clipping that exists outside them.

I have implemented this by storing the overall clip region outside the current container (`graphics->previous_clip`, `NULL` if not in a container or if no clipping outside) in addition to any clip region set (`graphics->clip`), and then intersecting them in `cairo_SetGraphicsClip()`, `GdipBeginContainer2()` and `GdipGetVisibleClipBounds()` (the three places that need the overall clip). `previous_clip` goes on the save stack on saving, in addition to `clip`.

Alternatively the intersection could be done when modifying the clip and when restoring it off the save stack (four places), but that would mean storing the resulting region as well as the other two, although the overall clip wouldn't need to be stored on the save stack. (`GdipTranslateClip()`, `GdipGetClip()` and `GdipGetClipBounds()` need the clip that was set.)

Also, I modified `GdipResetClip()` to call `cairo_SetGraphicsClip()` rather than `cairo_ResetClip()` when `previous_clip` exists, as it needs to keep that. Alternatively, it would be possible to modify `cairo_ResetClip()` to set the clip to `previous_clip` itself.

I have implemented this into `GdipGetVisibleClipBounds()`, as that should provide the overall clip bounds. And I changed `GdipIsVisiblePoint()` and `GdipIsVisibleRect()` to use that instead of the `graphics->bounds` (they should take clipping into account). I won't mention the insane check previously used in the latter of those two methods, but the replacement was taken from region.c – ideally I'd have added a declaration for it to region-private.h, but I didn't think `gdip_intersects()` was a descriptive enough name for that, and I didn't want to do a rename.

I guess I should write some tests too, so I'll try to get around to that some time soon. But I'll put this out there now for any comments.

### DIFF
--- a/src/graphics.c
+++ b/src/graphics.c
@@ -2451,8 +2451,7 @@ GdipIsVisiblePointI (GpGraphics *graphics, INT x, INT y, BOOL *result)
 GpStatus WINGDIPAPI
 GdipIsVisibleRect (GpGraphics *graphics, REAL x, REAL y, REAL width, REAL height, BOOL *result)
 {
-	float posy, posx;
-	GpRectF recthit, boundsF;
+	GpRectF boundsF;
 
 	if (!graphics || !result)
 		return InvalidParameter;
--- a/src/graphics.c
+++ b/src/graphics.c
@@ -144,6 +144,7 @@ gdip_graphics_common_init (GpGraphics *graphics)
 
 	GdipCreateRegion (&graphics->clip);
 	GdipCreateMatrix (&graphics->clip_matrix);
+	graphics->previous_clip = NULL;
 	graphics->bounds.X = graphics->bounds.Y = graphics->bounds.Width = graphics->bounds.Height = 0;
 	graphics->orig_bounds.X = graphics->orig_bounds.Y = graphics->orig_bounds.Width = graphics->orig_bounds.Height = 0;
 	graphics->last_pen = NULL;
@@ -419,6 +420,11 @@ GdipDeleteGraphics (GpGraphics *graphics)
 		graphics->clip = NULL;
 	}
 
+	if (graphics->previous_clip) {
+		GdipDeleteRegion (graphics->previous_clip);
+		graphics->previous_clip = NULL;
+	}
+
 	if (graphics->clip_matrix) {
 		GdipDeleteMatrix (graphics->clip_matrix);
 		graphics->clip_matrix = NULL;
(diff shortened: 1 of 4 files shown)


PreferLinux:
I realised shortly after I posted this yesterday that `GdipIsVisiblePoint()` and `GdipIsVisibleRect()` should not be using the clip _bounds_, they should be using the clip itself... So I'll sort that out. I'll also change to the alternative I mentioned above, because that'll result in less intersecting of `clip` with `previous_clip`.

filipnavara:
Sorry it is taking me so long to review this. I'm still going through the ASAN reports (after applying @akoeplinger's test patches). There are some leaks reported and I have to figure out whether they are false positives, pre-existing leaks, or something introduced by this PR (likely mix of all of this).

src/graphics.c
```diff
@@ -516,14 +531,29 @@ GdipRestoreGraphics (GpGraphics *graphics, GraphicsState state)
...
+	}
+	if (pos_state->previous_clip) {
+		status = GdipCloneRegion (pos_state->previous_clip, &graphics->previous_clip);
+		if (status != Ok)
+			return status;
+	}
+
 	gdip_cairo_matrix_copy (graphics->clip_matrix, &pos_state->clip_matrix);
 
+	gdip_calculate_overall_clipping (graphics);
```
filipnavara, line 555: `graphics->overall_clip` is leaked at this point. Couple of lines above you set `graphics->previous_clip` to `NULL`. This causes `gdip_calculate_overall_clipping` to enter this code path:

```
	if (!graphics->previous_clip) {
		graphics->overall_clip = graphics->clip;
	}
```

and never free the original `overall_clip`.

src/graphics.c
```diff
@@ -516,14 +531,29 @@ GdipRestoreGraphics (GpGraphics *graphics, GraphicsState state)
 
 	GdipSetRenderingOrigin (graphics, pos_state->org_x, pos_state->org_y);
 
-	if (graphics->clip)
+	if (graphics->clip) {
+		if (graphics->overall_clip == graphics->clip)
+			graphics->overall_clip = NULL; // Don't double-free!
 		GdipDeleteRegion (graphics->clip);
+	}
```
filipnavara, line 538:
```suggestion
	if (graphics->overall_clip != graphics->clip) {
		GdipDeleteRegion (graphics->overall_clip);
	}
	graphics->overall_clip = NULL;

	if (graphics->clip) {
		GdipDeleteRegion (graphics->clip);
	}
```

PreferLinux:
Thanks for picking that up. Your suggested fix looks good to me, so I've committed it.

PR #641, opened 2020-04-07
Fix comparison for color keying when applying image attributes

filipnavara:
Fixes https://github.com/mono/mono/issues/8514

### DIFF
--- a/src/imageattributes.c
+++ b/src/imageattributes.c
@@ -349,7 +349,7 @@ gdip_process_bitmap_attributes (GpBitmap *bitmap, void **dest, GpImageAttributes
 				GdipBitmapGetPixel (bmpdest, x, y, &color);
 				color &= ~ALPHA_MASK;
 				
-				if (color >= trans->key_colorlow && color <= trans->key_colorhigh) {
+				if (color >= (trans->key_colorlow & ~ALPHA_MASK) && color <= (trans->key_colorhigh & ~ALPHA_MASK)) {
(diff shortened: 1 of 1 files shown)


sancheolz:
Looks better. Now the icons have a white background. In Windows background transparent
Here is an example toolbar from windows
winToolbar
Here mono master & libgdiplus master
monoMasterGDI
Here mono master & libgdiplus master with this patch
monoPatchedGDI
Here res file with fifth icon in the last screenshot
[Plans.zip](https://github.com/mono/libgdiplus/files/4448928/Plans.zip)
I get it from the resource with the following code
```cs
                    Bitmap bmp =
                    new Bitmap(
                        Assembly.GetExecutingAssembly().GetManifestResourceStream(
                            "MyProject.Images.Plans.bmp"));
                var rect = new Rectangle(0, 0, bmp.Height, bmp.Height);
                return bmp.Clone(rect, bmp.PixelFormat);
```
It has Color depth = 4
The fourth icon in the last screenshot has a color depth = 8
both bmp format icons

filipnavara:
@sancheolz It could be related to the FIXME in the code:
https://github.com/mono/libgdiplus/blob/8411dce64028204a84d16dc001e6c057fa5ffb3d/src/imageattributes.c#L341-L344
or entirely different issue altogether.

These code paths definitely needs more testing and test coverage. This particular PR fixes just a regression from PR #567.

PR #646, opened 2020-04-17
Fix System.OutOfMemoryException when 512 calls to BeginContainer were made since application start

n432sean:
Fixing an issue with the state count not being decremented (stack not popped) when EndContainer is called. This would cause a System.OutOfMemoryException when 512 calls to BeginContainer were made since application start.


filipnavara:
It was broken by #561.

akoeplinger:
Cherry-picked to release/6.0.

PR #647, opened 2020-04-19
Fix double free SIGABRT with custom dash array handling

cpw:
Fix double free crash if you assign a custom dash array then switch back to a standard pattern.

### DIFF
--- a/src/pen.c
+++ b/src/pen.c
@@ -821,26 +821,31 @@ GdipSetPenDashStyle (GpPen *pen, GpDashStyle dashstyle)
 	case DashStyleSolid:
 		pen->dash_array = NULL;
 		pen->dash_count = 0;
+		pen->own_dash_array = FALSE;
 		break;
 
 	case DashStyleDashDot:
 		pen->dash_array = DashDot;
 		pen->dash_count = 4;
+		pen->own_dash_array = FALSE;
 		break;
 
 	case DashStyleDashDotDot:
 		pen->dash_array = DashDotDot;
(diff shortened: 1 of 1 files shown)


filipnavara:
This actually made me think. Should not the `pen->dash_array` be freed if it was previously set to custom value? What is the expected behavior when switching to standard pattern and then back to custom pattern using `GdipSetPenDashStyle`?

akoeplinger:
>Should not the pen->dash_array be freed if it was previously set to custom value? 

We seem to be `memcpy` the custom dash array here so yeah I guess it makes sense that we'd free it: https://github.com/mono/libgdiplus/blob/2d10e5d0c44c5f7ce55656d4398bbe5825033fff/src/pen.c#L941

> What is the expected behavior when switching to standard pattern and then back to custom pattern using GdipSetPenDashStyle?

I don't know and couldn't find any docs about that. If it was expected that switching back and forth retained the custom pattern then we shouldn't free.

filipnavara:
> We seem to be memcpy the custom dash array here so yeah I guess it makes sense that we'd free it:

That means it would currently leak in the previous value in this method when switching from the custom pattern to a predefined one. I can submit PR to fix that.

> If it was expected that switching back and forth retained the custom pattern then we shouldn't free.

The custom pattern gets overwritten by this method so the switch back would not work. I don't know if it should work though. Wine seems to retain the custom pattern but it's only consequence of different implementation.

The CI seemed to hard fail on this PR with a memory corruption error. I didn't look into it yet.

PR #648, opened 2020-04-27
Fix memory leak when setting pen dash style from custom to predefined

filipnavara:
This fixes part of #647 that I complained about in comment.

I tested locally that Windows libgdiplus in fact resets the dash array on `DashStyleCustom` -> `DashStyleDashDot` -> `DashStyleCustom` transition done using `GdipSetPenDashStyle`. Following this transition `GdipGetPenDashArray` returns the Dash-Dot pattern and not the original custom one.

### DIFF
--- a/src/pen.c
+++ b/src/pen.c
@@ -817,6 +817,13 @@ GdipSetPenDashStyle (GpPen *pen, GpDashStyle dashstyle)
 	if (!pen)
 		return InvalidParameter;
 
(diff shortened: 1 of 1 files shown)


akoeplinger:
The other pending builds actually finished and are green on Azure DevOps: https://dev.azure.com/dnceng/public/_build/results?buildId=618761&view=results

PR #650, opened 2020-04-28
Various memory leak fixes

filipnavara:
This builds up on fixes by @akoeplinger and fixes remaining memory leak errors reported by the ASAN/LSAN tools. It passes the included test suite (`make check`) and gtest suite (`make -C tests run-gtest`) on fully updated Ubuntu Linux 1804.

### DIFF
--- a/src/image.c
+++ b/src/image.c
@@ -663,6 +663,11 @@ GdipDrawImageRectRect (GpGraphics *graphics, GpImage *image,
 		srcheight = gdip_unit_conversion (srcUnit, UnitCairoPoint, graphics->dpi_y, graphics->type, srcheight);
 	}
 
+	if (gdip_near_zero (srcwidth) || gdip_near_zero (dstwidth) ||
+	    gdip_near_zero (srcheight) || gdip_near_zero (dstheight)) {
+		return Ok;
+	}
+
 	org = dest = image->active_bitmap->scan0; 
 	org_format = image->active_bitmap->pixel_format;
 	org_surface = image->surface;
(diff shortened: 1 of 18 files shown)


src/image.c
```diff
@@ -863,7 +870,7 @@ GdipDrawImageRectRect (GpGraphics *graphics, GpImage *image,
 		image->active_bitmap->scan0 = org;
 		image->active_bitmap->pixel_format = org_format;
 		image->surface = org_surface;
-		// NOTE: dest is freed by gdip_bitmap_invalidate_surface above
+		GdipFree (dest);
```
filipnavara, line 873: If we still want to backport anything to the 6.0 branch then this would be the one line. I am planning to restructure the code in a way that will make it much less error prone.

akoeplinger: how critical do you think this is? I'd like to avoid another 6.0 release

tests/testfont.c
```diff
@@ -225,6 +225,10 @@ static void test_privateAddFontFile ()
 	assertEqualInt (status, Ok);
 	assert (stringsEqual (name, "Code New Roman"));
 
+	// This causes a crash in GDI+.
+#if !defined(USE_WINDOWS_GDIPLUS)
+	GdipDeleteFontFamily (families[0]);
+#endif
```
filipnavara, line 231: This and the three remaining identical block below are not technically correct. `GdipGetFontCollectionFamilyList` should not return duplicates of the font family objects. However, modifying the test makes checking with ASAN/LSAN easier. If it ever gets fixed we can fix it in a way that these calls won't crash. Attempt to explicitly delete font family objects that are part of font collection can be treated as no-op.

src/image.c
filipnavara: I am fine with avoiding it since I expect this to be the last big missing piece for 6.1 release. It's a bigger leak if the code path is hit though.

akoeplinger: Ok let's not backport it then.

PR #654, opened 2020-06-26
Implement CachedBitmap functionality

reflectronic:
This PR implements the [CachedBitmap](https://docs.microsoft.com/en-us/windows/win32/gdiplus/-gdiplus-cachedbitmap-flat) functionality from GDI+. This is in preparation for implementing https://github.com/dotnet/runtime/issues/36105, which adds `CachedBitmap` to `System.Drawing`.

I would like to get CachedBitmap implemented for .NET 5. There are only 19 or so days until the window closes. @filipnavara, hopefully I can get this quickly reviewed, ironed out, and merged, so that I can finalize the .NET implementation.

.NET needs to be able to fail gracefully if the installed version of libgdiplus does not support CachedBitmap. I understand there is a facility for checking the version — `GetLibgdiplusVersion()` — but when will this version number be next incremented? Naturally, if using this is not appropriate, then we can just probe for the proper exports on the managed side.


src/cachedbitmap.c
```diff
@@ -0,0 +1,56 @@
...
+
+	status = cairo_surface_status (surface);
+	if (status != CAIRO_STATUS_SUCCESS) {
+		cairo_surface_destroy (surface);
+		return gdip_get_status (status);
+	}
+
+	newCachedBitmap = GdipAlloc (sizeof (GpCachedBitmap));
+	if (!newCachedBitmap) 
+		return OutOfMemory;
```
filipnavara, line 38: Missing `cairo_surface_destroy (surface);` in this code path.

tests/testcachedbitmap.c
```diff
@@ -0,0 +1,131 @@
...
+using namespace Gdiplus;
+using namespace DllExports;
+#endif
+
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include "testhelpers.h"
+
+#define C(func) assert (func == Ok)
```
filipnavara, line 26: This definition doesn't match the usage later (1 parameter here, 2 parameters later). It would be nice to stick to the same test macros as in the other test files.

reflectronic: Must have made a mistake when fixing the formatting. As for the macro itself, I copied it straight out of https://github.com/mono/libgdiplus/blob/9efff6a9ed1c3286362270b704dcdcf7a2670d54/tests/testclip.c#L16

reflectronic:
Ah, I forgot about license headers. I'm not really sure what to do with them, since the ones in other files are ancient. Should I use the .NET Foundation license headers found in other projects?

tests/testcachedbitmap.c
```diff
@@ -0,0 +1,130 @@
...
+
+	C (GdipDeleteGraphics(graphics));
+	C (GdipDisposeImage(image));
+}
+
+int main (int argc, char**argv)
+{
+	STARTUP;
+
+	test_create ();
```
hughbe, line 125: can we add tests for `test_delete`

reflectronic:
@akoeplinger if you could take a look and merge this soon, that would be great. I hope to get this in soon, so that .NET 5 can depend on it. Is there a time frame for when 6.1 will be releasing?

qmfrederik:
@reflectronic Really great to see new features being implemented in libgdiplus!

I went through the process of getting .NET Core code to depend on a new version of libgdiplus.
- On Linux, the current policy of the .NET Core team is to test against the version of libgdiplus which ships with the various Linux distributions (as opposed to the latest version of libgdiplus from the Mono repositories). That means you'll need to get a release in libgdiplus, get someone to upstream those releases to the distros, and wait for a new distro release.
- On macOS, that same policy implies .NET Core relies on whatever version ships with Homebrew. All it takes for Homebrew to ship a new version of libgdiplus is a tag in this repo. You can then open a PR against Homebrew and have it merged fairly quickly. However, as a next step, you'll need someone to run `brew update libgdiplus` the .NET Core build machines. Last time, that process took 5 months: dotnet/corefx#41707 dotnet/runtime#335

Let's hope the process has been ironed out a bit by now :).

On the other hand, System.Drawing.Common ships as a NuGet package, so it's a bit easier to consume newer versions of System.Drawing.Common.

reflectronic:
Yeah, surely the process of getting a new version shipped is going to take a while. Since this feature introduces new API surface (possibly for the first time in a long while), I implemented [version detection](https://github.com/reflectronic/runtime/blob/cachedbitmap/src/libraries/System.Drawing.Common/src/System/Drawing/Imaging/CachedBitmap.Unix.cs#L8-L19) so that the managed API can fail gracefully on an older version.

As for running tests in dotnet/runtime, currently my plan was to just skip the tests for now; when 6.1 eventually makes its way to the build machines, the tests would just light up. When I submit my PR to the runtime repository, then I could probably get some guidance from the area owners on what to do. Hopefully something can be figured out.

reflectronic:
For the record, I don't think it's critical to get this shipped everywhere within the next 5 months. For example, one of the biggest scenarios for System.Drawing is Windows Forms; for those who weren't going to rely on libgdiplus anyway, adding CachedBitmap to .NET 5 is going to work flawlessly. Whenever 6.1 does come out, then CachedBitmap can light up on non-Windows platforms from the managed side. However, at the least, we do need to merge this first, so that I can ensure that whatever *does* ship in the future will be compatible with the way it's implemented in .NET 5.

danmosemsft:
@akoeplinger, @reflectronic has a matching PR for this open against dotnet/runtime. Do you expect to merge this PR? I guess even when you do, any code in System.Drawing would have to be "light up" for some time.

PR #670, opened 2020-09-08
Vertical text alignment and positioning fix

PreferLinux:
Before fix:
Screenshot_20200908_152640
The vertical text is being drawn off-screen. Note the boxes (from measuring the string) for vertical left-to-right are correct, hence no test failures. `GdipAddPathString()` has much the same issue, so I've added a test there that has vertical text to detect the incorrect drawing location (fails without fix). I've also added tests for alignment when drawing at a point instead of in a box, and drawing with both vertical and right-to-left (which fail without the fix).

Here's a screenshot with only the changes to `pango_DrawString` applied, fixing the drawing location:
Screenshot_20200908_154720
On vertical text the alignment is opposite what it should be, and on right-to-left it is also significantly offset horizontally. Hence there are also changes dealing with alignment.

With all changes:
Screenshot_20200908_152509
.Net / Windows 8.1 as a comparison:
screenshot-Windows-text-alignment

Test application: [text-align.txt](https://github.com/mono/libgdiplus/files/5185921/text-align.txt)
Compiles with `mcs -r:System.Windows.Forms,System.Drawing text-align.txt -unsafe`

Furthermore, as `gdip_pango_setup_layout` transforms the Cairo context on vertical text, `GdipAddPathString()` wasn't getting the correct path points. I've changed it to save and later restore the context to fix this.

### DIFF
--- a/src/graphics-path.c
+++ b/src/graphics-path.c
@@ -1296,10 +1296,24 @@ GdipAddPathString (GpPath *path, GDIPCONST WCHAR *string, int length,
 		return status;
 	}
 
+	cairo_save(cr); // gdip_pango_setup_layout can transform the Cairo context (vertical text), so save it and restore later.
+
 	layout = gdip_pango_setup_layout (cr, string, length, font, layoutRect, &box, &box_offset, string_format, NULL);
-	cairo_move_to (cr, layoutRect->X + box_offset.X, layoutRect->Y + box_offset.Y);
+	if (string_format->formatFlags & StringFormatFlagsDirectionVertical) {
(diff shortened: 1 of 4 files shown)


PreferLinux:
Nice, a test failure in CI that I don't get on my machine. Will have to add some debugging to that test, and then revert it...

PreferLinux:
Actually, didn't need to. It is simply that the Linux testing uses Cairo text rendering (due to an old Pango version), and that can't handle the tests I added. Those tests have now been made conditional.

PR #671, opened 2020-10-02
Don't dispose of GpMetafile if a GpGraphics is using it

lambdageek:
Allow GpGraphics to take ownership of GpMetafile on dispose

The Windows GDI+ allows the metafile instance to be disposed before the graphics instance.

To support that, keep track in the GpGraphics of whether it is responsible for freeing the GpMetafile.

If the graphics instance is deleted first, own_metafile is FALSE and nothing happens.

If the metafile instance is disposed first, set own_metafile to TRUE and don't dispose of the metafile. It will be disposed when the graphics instance is deleted.

---

Also on Windows GDI+ creating more than one graphics instance for a metafile
instance throws an OutOfMemoryException, so do the same in GdipGetImageGraphicsContext

---

This is related to https://github.com/dotnet/runtime/issues/37838

### DIFF
--- a/src/graphics-private.h
+++ b/src/graphics-private.h
@@ -126,6 +126,7 @@ typedef struct _Graphics {
 	float			aa_offset_y;
 	/* metafile-specific stuff */
 	EmfType			emf_type;
+	BOOL			own_metafile;	/* true when metafile is owned by the graphics instance */
 	GpMetafile		*metafile;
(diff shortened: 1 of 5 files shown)


lambdageek:
Looks like on Win32 GDI+, you can dispose of the metadata or the graphics instances in either order. Instead of asserting, I'm working on extending the lifetime of the metadata object until the graphics instance is disposed if the metadata is disposed first.

PR #674, opened 2020-12-04
Add unit test for GdipDrawImageRectRectI

qmfrederik:
This adds a unit test for `GdipDrawImageRectRectI`. It will resize a white rectangle and draw it on top of a black image, and asserts that the result is (more or less) white.

### DIFF
--- a/tests/gtest.cpp
+++ b/tests/gtest.cpp
@@ -106,6 +106,67 @@ TEST(GraphicsTest, FromHicon_InvalidHandle_ThrowsArgumentException) {
 }
 #endif
 
+TEST(GraphicsTest, DrawImage_Test) {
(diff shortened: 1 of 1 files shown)


qmfrederik:
I created this test because some of my code started failing when being used with libgdiplus 6.0.4 (which is the default on Ubuntu 20.04 and later).

The libgdiplus 6.0 behavior differs from what GDI+ does (and libgdiplus < 6.0) when you use `PixelFormat32bppARGB`. I didn't dig deeper into that, as the images which I'm drawing don't have alpha channels and the fix was to just use the RGB pixel format instead of ARGB.

But I did came up with this unit test, and thought it'd be a good addition to the test suite, anyway ;-).

PR #676, opened 2021-01-03
[Region Refactoring] Refactor gdip_region_convert_from_path

hughbe:
- Clear the region in `gdip_region_create_from_path`
- Add missing status checks in calls to `GdipAddPathRectangle`
- Explicitly handle `RegionTypeInfinite` in `gdip_region_convert_to_path`
- Implement `gdip_region_convert_to_path` on top of `gdip_region_create_from_path`
- Rename `gdip_region_create_from_path` to match meaning

This change achieves some things
- Adds missing status checks and memory leaks in the error path where GdipAddPathRectangle fails
- Avoid duplicate code (e.g. creating the path tree and clearing the existing path data) between `gdip_region_convert_from_path` and `gdip_region_create_from_path`
- Avoid duplicate code clearing the path data each time we call `gdip_region_create_from_path`: we should always clear existing data!
- Rename `gdip_region_create_from_path` to `gdip_region_set_path` to better match the meaning of the method - we're not actually creating a region, but rather overwriting existing data

The motivation for this is to avoid duplicate code and enable future refactorings such as lazily creating path tree data

Should be merged BEFORE #675

### DIFF
--- a/src/graphics.c
+++ b/src/graphics.c
@@ -2289,8 +2289,7 @@ GdipGetClip (GpGraphics *graphics, GpRegion *region)
 	if (graphics->state == GraphicsStateBusy)
 		return ObjectBusy;
 
-	gdip_clear_region (region);
-	gdip_copy_region (graphics->clip, region);
+	gdip_region_set_region (region, graphics->clip);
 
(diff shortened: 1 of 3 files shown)


hughbe:

Issue #338, opened 2018-04-13
Intersections involving infinite paths fail with OutOfMemory


hughbe:
We fail to allocate the bitmap

PR #678, opened 2021-01-04
Add more GdipGetRegionData tests and fix GdipCombineRegionRect for infinite rects

### LINKED ISSUE #339: Union of an empty region with an infinite region should be infinite, not empty

hughbe:
- Fixes #339 by correctly handling infinite and empty rects to match GDI+ behaviour
- Contributes to #338 by fixing `region (Exclude) infinite region` to always set to empty
- Add more `GdipGetRegionData` that cover this bug. Note that many of them are currently disabled as they pass on GDI+ but not libgdiplus. More work to be done.


hughbe:

PR #689, opened 2021-01-25
Set pango dependency as required

### LINKED ISSUE #688: Next release with required pango dependency by default
Most libgdiplus packages are built without `--with-pango` and .NET users are forced to build the library from source to get the advanced typography features working. This need for building from source is not obvious to the end users (there are multiple issues in dotnet/runtime repo attesting to that) or the package maintainers.

Back in 2018, some package management systems were having `pango < v1.38` and therefore marking the pango dependency mandatory for libgdiplus (which at the time worked with `pango >= v1.38`) was not feasible:  https://github.com/mono/libgdiplus/pull/269#issuecomment-378706342.

However, now the situation (based on my research) is so; all relevant packaging systems with libgdiplus v6x also have `pango > 1.40`. Distros, such as Ubuntu 18.04, which have older pango are also <ins>not</ins> upgrading libgdiplus for some time (bionic is still has libgdiplus v4x). Later versions of Ubuntu have both pango 1.4x and libgdiplus v6x. Therefore marking pango as required dependency is safe.

Suggestion for next libgdiplus version:
* remove `--with-pango` argument.
* error when `--with-pango` argument is provided during configure: `` Remove `--with-pango` as pango is now a required dependency.``
* add `--without-pango` argument with a `` *WARNING* - some typrography / text rendering features will not work without pango. Consider removing `--without-pango` and adding the dependency.``

am11:
Fixed #688

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -14,7 +14,7 @@ stages:
       displayName: Linux x64
       timeoutInMinutes: 120
       pool:
-        vmImage: ubuntu-16.04
+        vmImage: ubuntu-20.04
       steps:
         - bash: |
             sudo apt update
@@ -165,7 +165,7 @@ stages:
       displayName: Linux x64 mono-netcore
       timeoutInMinutes: 120
       pool:
-        vmImage: ubuntu-16.04
+        vmImage: ubuntu-20.04
       steps:
         - bash: |
             sudo apt update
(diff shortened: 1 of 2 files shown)


am11:
@akoeplinger, it would be great if we get a new release tag (6.0.6 or 6.1.0) after this PR is merged. There are plenty of @filipnavara's memory leak fixes since 6.0.5, which downstream packages are missing. 😁

am11:
Seems like CI jobs are stuck in PRs since 10.2020. Maybe we just need to revoke the status check to get a clean state?
* Push a commit to master deleting AZDO yml.
* Clear status check rules from settings/branches.
* Push the AZDO yaml file back to master branch.
  * Double-check the status checks.

Stuck PRs can then rebase against the master branch.

PR #691, opened 2021-02-08
Keeping track of temporary files created by GdipPrivateAddMemoryFont

### LINKED ISSUE #690: GdipPrivateAddMemoryFont does not clean up temporary files
Trying to use a C# application that calls `System.PrivateFontCollection.AddMemoryFont` on Linux leaves a bunch of unused font files in the `tmp` folder, even after calling `Dispose`.

Most functions in Mono's PrivateFontCollection are simply wrappers to `libgdiplus` calls, so I guess the bug originates from here.

Digging through the source code, it seems that this piece of code [here](https://github.com/mono/libgdiplus/blob/9efff6a9ed1c3286362270b704dcdcf7a2670d54/src/font.c#L1414) is responsible:
```
#ifdef WIN32
	f = CreateTempFile (fontfile);
	if (!f)
		return FileNotFound;

	if (fwrite(memory, sizeof(BYTE), length, f) != length) {
		fclose (f);
		return FileNotFound;
	}

	fclose (f);
#else
	strcpy ((char *) fontfile, "/tmp/ffXXXXXX");
	f = mkstemp ((char *) fontfile);
	
	if (f == -1)
		return FileNotFound;

	if (write (f, memory, length) != length) {
		close (f);
		return FileNotFound;
	}
	close (f);
#endif

	FcConfigAppFontAddFile (fontCollection->config, fontfile);
	/* FIXME - May we delete our temporary font file or does 
	   FcConfigAppFontAddFile just reference our file?  */
	/* unlink(fontfile); */

	return Ok;
```

After research, at least on my Ubuntu 20.04 machine, `FcConfigAppFontAddFile` simply references that file, deleting the font makes the app crash when a new Form or Control is drawn. However, the function does not seem to save the temp files anywhere.

trungnt2910:
This should fix #690.

I've added a list of strings that should go along with the `GpFontCollection`, this list are the temporary files created by the `GdipPrivateAddMemoryFont` function.

When `Dispose` is called by C# code, `GdipDeletePrivateFontCollection` should go through the list and `remove`s all temp files created.

There are ways to get files associated with an `FcConfig`, however, it will be hard to distinguish between files added using `GdipPrivateAddMemoryFont` or files managed by the user added through `GdipPrivateAddFontFile`, so I guess keeping a list of strings for later deletion is still a good solution.

### DIFF
--- a/src/font.c
+++ b/src/font.c
@@ -162,6 +162,36 @@ GdipNewInstalledFontCollection (GpFontCollection **fontCollection)
 	return Ok;
 }
 
+void free_tempfiles_list(StringList* head)
+{
+	while (head)
+	{
+		remove (head->str);
+		head->str = NULL;
+		head = head->next;
+	}
+}
+
+void destroy_string_list(StringList* head)
+{
+	while (head)
+	{
+		StringList* next = head->next;
+		GdipFree (head);
+		head = next;
+	}
+}
+
+GpStatus push_front_string_list(StringList** head, char* str)
+{
+	StringList* newHead = GdipAlloc (sizeof(StringList));
+	if (!newHead) return OutOfMemory;
+	newHead->str = str;
+	newHead->next = *head;
+	*head = newHead;
+	return Ok;
+}
+
 // coverity[+alloc : arg-*0]
 GpStatus WINGDIPAPI
 GdipNewPrivateFontCollection (GpFontCollection **fontCollection)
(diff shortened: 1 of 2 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/mono/libgdiplus?pullRequest=691) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/mono/libgdiplus?pullRequest=691) before we can accept your contribution.<br/><br/>:x: trungnt2910 [sign now](https://cla.dotnetfoundation.org/mono/libgdiplus?pullRequest=691)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/mono/libgdiplus?pullRequest=691) it.</sub>

PR #697, opened 2021-03-31
Big-endian fix: Header conversion

uweigand:
On big-endian systems, various file header elements need to be
byte-swapped.  In a few places, current code used the conversion
function for the wrong data type.


uweigand:
Ping?

akoeplinger (approved):

PR #698, opened 2021-03-31
Big-endian fix: Accessing bitmaps in region-bitmap.c

uweigand:
The bitmaps used in region-bitmap.c are created by Cairo in
the the CAIRO_FORMAT_A1 format.  In this format, the order of
1-bit pixels stored in a byte varies depending on endianness.

Fix all accesses to these bits by introducing an is_bit_set
helper macro that takes endianness into account.


uweigand:
Ping?

akoeplinger (approved):

PR #699, opened 2021-03-31
Big-endian fix: Reading BMP files

uweigand:
When reading BMP files in 16bppRGB555 or 16bppRGB565 format,
the 16-bit container holding the RGB bits has to be read in
little-endian, even on big-endian systems.


uweigand:
Ping?

akoeplinger (approved):

PR #700, opened 2021-03-31
Big-endian fix: Reading PNG files

uweigand:
When reading PNG files, the result must match the Cairo format
CAIRO_FORMAT_ARGB32.  This stores the pixel data as a 32-bit
quantity in native byte order.

Therefore, the PNG decoder must place the bytes in A,R,G,B
order on a big-endian system (as opposed to B,G,R,A on a
little-endian system).  Set up the libpng flags accordingly.


uweigand:
Ping?

akoeplinger (approved):

PR #701, opened 2021-04-10
Build unit tests only when --enable-unittests is passed

sergioprado:
Building unit tests by default requires all dependencies to be
installed (jpeg, tiff, etc) and users may not want that.

So let's disable the build of unit tests by default and introduce
--enable-unittests configure option to enable it.

Signed-off-by: Sergio Prado <sergio.prado@e-labworks.com>


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/mono/libgdiplus?pullRequest=701) <br/>All CLA requirements met.

